PR #1, opened 2019-10-30
PARQUET-1674: The announcement email on the web site does not comply with ASF rules

gszadovszky:
Apache requires to have our own download page that contains links to the source tarball, the related ASC and SHA512 files, the KEYS file and instructions for verifying the downloaded file. These links shall not be included in the announcement email but the link of the download page only.


rdblue:
Can you please update the description with what the policy is and what needed to be changed?

gszadovszky:
Added some notes about the apache requirements. More precise requirements and the related links are available in the jira.

gszadovszky:
@rdblue, Could you please take a look?

rdblue:
+1

PR #2, opened 2019-11-05
PARQUET-1687: Update release process

gszadovszky:
Update the link of the keys file to the official one. The official link
is required by the Apache release process.
Update the usage of the prepare script to use the RC number as a
parameter so they can create RC tags.
Add usage of the new finalize script that creates the final release tag
and sets the new development version.
As we are using RC tags and add the final release tag only after the
vote passes we do not need to do anything in case of the vote fails.

### DIFF
--- a/source/documentation/how-to-release.html.md
+++ b/source/documentation/how-to-release.html.md
@@ -4,7 +4,7 @@
 ### Setup
 
 You will need:
-* PGP code signing keys, published in KEYS
+* PGP code signing keys, published in [KEYS][keys]
 * Permission to stage artifacts in Nexus
 
 Make sure you have permission to deploy Parquet artifacts to Nexus by pushing a snapshot:
@@ -15,6 +15,7 @@ mvn deploy
 
 If you have problems, read the [publishing Maven artifacts documentation][publish-maven-docs]
 
+[keys]: https://www.apache.org/dist/parquet/KEYS
 [publish-maven-docs]: https://www.apache.org/dev/publishing-maven-artifacts.html
 
 ### Release process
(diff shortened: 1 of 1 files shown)


source/documentation/how-to-release.html.md
```diff
@@ -68,7 +69,7 @@ Closing a staging repository makes the binaries available in [staging][staging],
 #### 4. Run the source tarball script
 
 ```
-sh dev/source-release.sh <version> <rc-number>
+dev/source-release.sh <version> <rc-number>
```
zivanfi, line 72: dev/source-release.sh is not executable in my repo, it doesn't work without `sh` or `.`

source/documentation/how-to-release.html.md
```diff
@@ -90,24 +91,23 @@ Hi everyone,
...
 Binary artifacts are staged in Nexus here:
-* https://repository.apache.org/content/groups/staging/org/apache/parquet/parquet/
+* https://repository.apache.org/content/groups/staging/org/apache/parquet/
 
 This release includes important changes that I should have summarized here, but I'm lazy.
 
 Please download, verify, and test.
 
-Please vote by <72 HOUR FROM NOW>
+Please vote in the next 72 hours.
```
zivanfi, line 110: I think we usually phrase it as "The vote will be open for at least 72 hours." so that we don't have to fail the release if people are slow to vote.

source/documentation/how-to-release.html.md
zivanfi: My bad, didn't notice the corresponding change in the other PR.

source/documentation/how-to-release.html.md
gszadovszky: I think that "Please vote in the next 72 hours" does not mean the vote will be closed after 72 hours. It only requests members to vote in 72 hours so the vote can be closed "soon".
I'm afraid the meaning of the suggested one might be "I don't have to hurry to vote because it won't be close even after 72 hours."

zivanfi (approved):

PR #7, opened 2022-01-05
source/documentation/latest: update links

kevinburkesegment:
The compatibility library has not been kept up to date and the Rust
library previously linked to a read-only repository. Update both.

I am using a M1 Mac and needed to update the ffi dependency in order
to get middleman to work correctly; I can revert that change or commit
it separately.


shangxinli (approved):

shangxinli:
@kevinburkesegment Thanks for working on it!

PR #29, opened 2022-09-18
Clarify Bit-Packing doc example positioning

Skwiwel:
When reading through the documentation I stumbled upon the Hybrid Bit-Packing section. The section is currently above the deprecated Bit-Packing encoding section, but refers to the example in the latter in a way that presumes the reader has already read it, which left me sightly confused.

Here's an attempt to better point the reader to the other example so that he can easier compare both methods.


pitrou (approved):

pitrou:
Sorry for the delay in taking a look at this extremely old PR.

@alamb Does it look ok to you too?

alamb (approved):
I agree -- this looks good to me. 

Thanks for the PR @Skwiwel and the review @pitrou

PR #34, opened 2023-06-20
PARQUET-2310: implementation status

alippai:
Moved from: https://github.com/apache/arrow/pull/36027

I only can agree with @westonpace:

>Although...to play devil's advocate...it might be odd when a feature is available in the parquet reader, but not yet exposed in the query component. For example, there is some row skipping and bloom filters in the C++ parquet reader, but we haven't integrated those into the datasets layer yet.

The original goal was not copying https://arrow.apache.org/docs/cpp/parquet.html#supported-parquet-features, but quite orthogonal to this: showing which _high level APIs_ are available for the end-user.

What I mean by this is that based on the document I should be able to know if I can manipulate Parquet files without decoding some values or do I have to convert it to actual objects / Apache Arrow table first. This was the info what is missing from all the docs and I'd like maintain something which reflects the public API functionalities.

Many of you were surprised by including xy metadata, what I was referring to was the availability of the https://arrow.apache.org/docs/python/generated/pyarrow.Schema.html#pyarrow.Schema.metadata or https://arrow.apache.org/docs/python/generated/pyarrow.Field.html#pyarrow.Field.metadata

Similar happens with the _Column Index_. It's available internally, but if it's not used when adding filters to the parquet reader, that's useful info for the Apache Arrow and Parquet developers, but not for the library users. 

This is a limitation of the pyarrow API and this might be temporary or by design, but making this transparent was my intention.

It's absolutely OK if parquet-mr doesn't have a dataset API using the parquet features and it's OK for pyarrow not exposing the raw objects / byte arrays (ever). Different ecosystems have different abstractions and there doesn't have to be a "correct" or "standard" one. This is actually a nice thing - DataFusion, pyarrow, Acero, Iceberg etc have all different communities complementing each other.

Let me know if this is too vague or doesn't belong to the standard docs (or help me achieving a concise summary reflecting the above)

### DIFF
--- a/content/en/docs/File Format/implementationstatus.md
+++ b/content/en/docs/File Format/implementationstatus.md
@@ -3,6 +3,26 @@ title: "Implementation status"
 linkTitle: "Implementation status"
 weight: 8
 ---
+
+This page summarizes the features supported by different Parquet
+implementations.
+
+*Note*: This is a work in progress and we would welcome help expanding its scope.
+
+### Legend
+The value in each box means:
+* ✅: supported
+* ❌: not supported
+* (blank) no data
+
+Implementations:
+* `C++`: [parquet-cpp](https://github.com/apache/arrow/tree/main/cpp/src/parquet)
+* `Java`: [parquet-java](https://github.com/apache/parquet-java)
+* `Go`: [parquet-go](https://github.com/apache/arrow/tree/main/go/parquet)
+* `Rust`: [parquet-rs](https://github.com/apache/arrow-rs/blob/master/parquet/README.md)
+
+
+
 ### Physical types
 
 | Data type                                 | C++   | Java   | Go    | Rust  |
--- a/content/en/docs/File Format/implementationstatus.md
+++ b/content/en/docs/File Format/implementationstatus.md
@@ -1,0 +1,101 @@
+---
+title: "Implementation status"
+linkTitle: "Implementation status"
+weight: 8
+---
+### Physical types
+
+| Data type                                 | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| BOOLEAN                                   |       |        |       |       |
+| INT32                                     |       |        |       |       |
+| INT64                                     |       |        |       |       |
+| INT96                                     |       |        |       |       |
+| FLOAT                                     |       |        |       |       |
+| DOUBLE                                    |       |        |       |       |
+| BYTE_ARRAY                                |       |        |       |       |
+| FIXED_LEN_BYTE_ARRAY                      |       |        |       |       |
+
+### Logical types
+
+| Data type                                 | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| STRING                                    |       |        |       |       |
+| ENUM                                      |       |        |       |       |
+| UUID                                      |       |        |       |       |
+| 8 and 16 bit signed INT                   |       |        |       |       |
+| 8, 16, 32, 64 bit unsigned INT            |       |        |       |       |
+| DECIMAL (INT32)                           |       |        |       |       |
+| DECIMAL (INT64)                           |       |        |       |       |
+| DECIMAL (BYTE_ARRAY)                      |       |        |       |       |
+| DECIMAL (FIXED_LEN_BYTE_ARRAY)            |       |        |       |       |
+| DATE                                      |       |        |       |       |
+| TIME (INT32)                              |       |        |       |       |
+| TIME (INT64)                              |       |        |       |       |
+| TIMESTAMP (INT32)                         |       |        |       |       |
+| TIMESTAMP (INT64)                         |       |        |       |       |
+| INTERVAL                                  |       |        |       |       |
+| JSON                                      |       |        |       |       |
+| BSON                                      |       |        |       |       |
+| LIST                                      |       |        |       |       |
+| MAP                                       |       |        |       |       |
+| UNKNOWN                                   |       |        |       |       |
+
+### Encoding
+
+| Encoding                                  | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| PLAIN                                     |       |        |       |       |
+| PLAIN_DICTIONARY                          |       |        |       |       |
+| RLE_DICTIONARY                            |       |        |       |       |
+| RLE                                       |       |        |       |       |
+| BIT_PACKED                                |       |        |       |       |
+| DELTA_BINARY_PACKED                       |       |        |       |       |
+| DELTA_LENGTH_BYTE_ARRAY                   |       |        |       |       |
+| DELTA_BYTE_ARRAY                          |       |        |       |       |
+| BYTE_STREAM_SPLIT                         |       |        |       |       |
+
+### Compression
+
+| Compression                               | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| UNCOMPRESSED                              |       |        |       |       |
+| SNAPPY                                    |       |        |       |       |
+| GZIP                                      |       |        |       |       |
+| LZO                                       |       |        |       |       |
+| BROTLI                                    |       |        |       |       |
+| LZ4                                       |       |        |       |       |
+| ZSTD                                      |       |        |       |       |
+| LZ4_RAW                                   |       |        |       |       |
+
+### Other format level features
+
+|                                           | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| xxHash Bloom filters                      |       |        |       |       |
+| bloom filter length                       |       |        |       |       |
+| Statistics min_value, max_value           |       |        |       |       |
+| Column index                              |       |        |       |       |
+| Offset index                              |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
+| Page CRC32 checksum                       |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
+
+### High level data API-s for parquet feature usage
+
+| Format                                       | C++   | Java   | Go    | Rust  |
+| -------------------------------------------- | ----- | ------ | ----- | ----- |
+| Hive-style partitioning                      |       |        |       |       |
+| Partition pruning on the partition column    |       |        |       |       |
+| External column data                         |       |        |       |       |
+| RowGroup Sorting column                      |       |        |       |       |
+| Read / Write RowGroup metadata and data (1)  |       |        |       |       |
+| RowGroup pruning using statistics            |       |        |       |       |
+| Read / Write page metadata and data (2)      |       |        |       |       |
+| Page pruning using projection pushdown       |       |        |       |       |
+| Page pruning using statistics                |       |        |       |       |
+| Page pruning using bloom filter              |       |        |       |       |
+
+* \(1) Ability to construct RowGroup objects from existing RowGroups or raw values (eg. without decoding or decompressing lower level data when read)
+
+* \(2) Ability to construct Page objects from existing Pages or raw values (eg. without decoding or decompressing lower level data when read)
--- a/content/en/docs/File Format/implementationstatus.md
+++ b/content/en/docs/File Format/implementationstatus.md
@@ -16,7 +16,7 @@ weight: 8
 | BYTE_ARRAY                                |       |        |       |       |
 | FIXED_LEN_BYTE_ARRAY                      |       |        |       |       |
 
-* \(1) This type is deprecated, but as of 2024 it's common in currently produced parquet files.
+* \(1) This type is deprecated, but as of 2024 it's common in currently produced parquet files
 
 
 ### Logical types
@@ -57,18 +57,18 @@ weight: 8
 | DELTA_BYTE_ARRAY                          |       |        |       |       |
 | BYTE_STREAM_SPLIT                         |       |        |       |       |
 
-### Compression
+### Compressions
 
 | Compression                               | C++   | Java   | Go    | Rust  |
 | ----------------------------------------- | ----- | ------ | ----- | ----- |
 | UNCOMPRESSED                              |       |        |       |       |
-| SNAPPY                                    |       |        |       |       |
-| GZIP                                      |       |        |       |       |
-| LZO                                       |       |        |       |       |
 | BROTLI                                    |       |        |       |       |
+| GZIP                                      |       |        |       |       |
 | LZ4 (deprecated)                          |       |        |       |       |
-| ZSTD                                      |       |        |       |       |
 | LZ4_RAW                                   |       |        |       |       |
+| LZO                                       |       |        |       |       |
+| SNAPPY                                    |       |        |       |       |
+| ZSTD                                      |       |        |       |       |
 
 ### Other format level features
 
--- a/content/en/docs/File Format/implementationstatus.md
+++ b/content/en/docs/File Format/implementationstatus.md
@@ -10,11 +10,14 @@ weight: 8
 | BOOLEAN                                   |       |        |       |       |
 | INT32                                     |       |        |       |       |
 | INT64                                     |       |        |       |       |
-| INT96                                     |       |        |       |       |
+| INT96 (1)                                 |       |        |       |       |
 | FLOAT                                     |       |        |       |       |
 | DOUBLE                                    |       |        |       |       |
 | BYTE_ARRAY                                |       |        |       |       |
 | FIXED_LEN_BYTE_ARRAY                      |       |        |       |       |
+
+* \(1) This type is deprecated, but as of 2024 it's common in currently produced parquet files.
+
 
 ### Logical types
 
@@ -23,8 +26,7 @@ weight: 8
 | STRING                                    |       |        |       |       |
 | ENUM                                      |       |        |       |       |
 | UUID                                      |       |        |       |       |
-| 8 and 16 bit signed INT                   |       |        |       |       |
-| 8, 16, 32, 64 bit unsigned INT            |       |        |       |       |
+| 8, 16, 32, 64 bit signed and unsigned INT |       |        |       |       |
 | DECIMAL (INT32)                           |       |        |       |       |
 | DECIMAL (INT64)                           |       |        |       |       |
 | DECIMAL (BYTE_ARRAY)                      |       |        |       |       |
@@ -32,16 +34,16 @@ weight: 8
 | DATE                                      |       |        |       |       |
 | TIME (INT32)                              |       |        |       |       |
 | TIME (INT64)                              |       |        |       |       |
-| TIMESTAMP (INT32)                         |       |        |       |       |
 | TIMESTAMP (INT64)                         |       |        |       |       |
 | INTERVAL                                  |       |        |       |       |
 | JSON                                      |       |        |       |       |
 | BSON                                      |       |        |       |       |
 | LIST                                      |       |        |       |       |
 | MAP                                       |       |        |       |       |
-| UNKNOWN                                   |       |        |       |       |
+| UNKNOWN (always null)                     |       |        |       |       |
+| FLOAT16                                   |       |        |       |       |
 
-### Encoding
+### Encodings
 
 | Encoding                                  | C++   | Java   | Go    | Rust  |
 | ----------------------------------------- | ----- | ------ | ----- | ----- |
@@ -49,7 +51,7 @@ weight: 8
 | PLAIN_DICTIONARY                          |       |        |       |       |
 | RLE_DICTIONARY                            |       |        |       |       |
 | RLE                                       |       |        |       |       |
-| BIT_PACKED                                |       |        |       |       |
+| BIT_PACKED (deprecated)                   |       |        |       |       |
 | DELTA_BINARY_PACKED                       |       |        |       |       |
 | DELTA_LENGTH_BYTE_ARRAY                   |       |        |       |       |
 | DELTA_BYTE_ARRAY                          |       |        |       |       |
@@ -64,7 +66,7 @@ weight: 8
 | GZIP                                      |       |        |       |       |
 | LZO                                       |       |        |       |       |
 | BROTLI                                    |       |        |       |       |
-| LZ4                                       |       |        |       |       |
+| LZ4 (deprecated)                          |       |        |       |       |
 | ZSTD                                      |       |        |       |       |
 | LZ4_RAW                                   |       |        |       |       |
 
@@ -72,30 +74,31 @@ weight: 8
 
 |                                           | C++   | Java   | Go    | Rust  |
 | ----------------------------------------- | ----- | ------ | ----- | ----- |
-| xxHash Bloom filters                      |       |        |       |       |
-| bloom filter length                       |       |        |       |       |
+| xxxHash-based bloom filters               |       |        |       |       |
+| Bloom filter length (1)                   |       |        |       |       |
 | Statistics min_value, max_value           |       |        |       |       |
-| Column index                              |       |        |       |       |
-| Offset index                              |       |        |       |       |
-| Modular encryption                        |       |        |       |       |
+| Page index                                |       |        |       |       |
 | Page CRC32 checksum                       |       |        |       |       |
 | Modular encryption                        |       |        |       |       |
+| Size statistics (2)                       |       |        |       |       |
 
-### High level data API-s for parquet feature usage
+
+* \(1) In parquet.thrift: ColumnMetaData->bloom_filter_length
+
+* \(2) In parquet.thrift: ColumnMetaData->size_statistics
+
+### High level data APIs for Parquet feature usage
 
 | Format                                       | C++   | Java   | Go    | Rust  |
 | -------------------------------------------- | ----- | ------ | ----- | ----- |
-| Hive-style partitioning                      |       |        |       |       |
-| Partition pruning on the partition column    |       |        |       |       |
-| External column data                         |       |        |       |       |
-| RowGroup Sorting column                      |       |        |       |       |
-| Read / Write RowGroup metadata and data (1)  |       |        |       |       |
-| RowGroup pruning using statistics            |       |        |       |       |
-| Read / Write page metadata and data (2)      |       |        |       |       |
-| Page pruning using projection pushdown       |       |        |       |       |
+| External column data (1)                     |       |        |       |       |
+| Row group "Sorting column" metadata (2)      |       |        |       |       |
+| Row group pruning using statistics           |       |        |       |       |
+| Reading select columns only                  |       |        |       |       |
 | Page pruning using statistics                |       |        |       |       |
 | Page pruning using bloom filter              |       |        |       |       |
 
-* \(1) Ability to construct RowGroup objects from existing RowGroups or raw values (eg. without decoding or decompressing lower level data when read)
 
-* \(2) Ability to construct Page objects from existing Pages or raw values (eg. without decoding or decompressing lower level data when read)
+* \(1) In parquet.thrift: ColumnChunk->file_path
+
+* \(2) In parquet.thrift: RowGroup->sorting_columns


content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
+---
+title: "Implementation status"
+linkTitle: "Implementation status"
+weight: 8
+---
+
+### Physical types
+
++-------------------------------------------+-------+--------+--------+-------+-------+
```
pitrou: Is this standard Markdown markup for tables? Github doesn't seem to recognize it (see what happens here if you ask to "view" the file).

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
+| Data type                                 | C++   | Python | Java   | Go    | Rust  |
```
pitrou: I think this should be a bit more specific than simply listing programming languages, as there might be several C++ or Java implementations? Replace them with their actual names.

Also, I don't think Python needs to be included here, since PyArrow is a binding around Arrow C++ (unless this is talking about another implementation, such as [fastparquet](https://fastparquet.readthedocs.io/en/latest/))?

content/en/docs/File Format/implementationstatus.md
alippai: I’ll check and convert it to the correct format. It was copy pasted from arrow-site

content/en/docs/File Format/implementationstatus.md
alippai: Agreed on replacing with actual names.

I wasn’t sure if currently or historically pyarrow features match arrow c++. Likely it can be omitted from some tables. I could imagine eg int96/nanos/decimal (fixed array) capabilities or compression options differing from c++ and on high level API the difference was already confusing for me.

pitrou: Yeah, PyArrow might not expose everything. But that's more of a PyArrow problem and I'm not sure it's worth exposing here. cc @jorisvandenbossche for opinions.

jorisvandenbossche: Looking at the list here, I think most of the features that C++ supports are also exposed in pyarrow, I think. 
Although I don't know if writing a type like UUID (that doesn't exist in arrow) is possible with the direct Parquet C++ API? (that's not possible through the Arrow based writer, I suppose, and so also not in pyarrow)

If we only keep the C++ column in the table, we should also add somewhere a note that this implementation is also exposed through Python, R, GLib, .. bindings. 
(another reason to not include Python, because then why not also include R, etc)

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
+
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Format                                       | C++   | Python | Java   | Go    | Rust  |
+|                                              |       |        |        |       |       |
++==============================================+=======+========+========+=======+=======+
+| Hive-style partitioning                      |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Partition pruning on the partition column    |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| External column data                         |       |        |        |       |       |
```
westonpace: What is this?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
++----------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup Sorting column                      |       |        |        |       |       |
```
westonpace: Also, what does this mean?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Read / Write RowGroup metadata and data (1)  |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup pruning using statistics            |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Read / Write page metadata and data (2)      |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using projection pushdown       |       |        |        |       |       |
```
westonpace: When I think "projection pushdown" I think "column selection".  Is that what is being discussed here?  I guess I wouldn't associate it with "page pruning" in my mind.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
+High level data API-s for parquet feature usage
+===============================================
+
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Format                                       | C++   | Python | Java   | Go    | Rust  |
+|                                              |       |        |        |       |       |
++==============================================+=======+========+========+=======+=======+
+| Hive-style partitioning                      |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Partition pruning on the partition column    |       |        |        |       |       |
```
westonpace: I think these two features could probably be left out.  I can see how the others are high-level API features but they are still very much "exposing parquet capabilities".  I think hive-style partitioning is completely unrelated to parquet the format though.

alippai: The number of systems which support parquet, but not hive style partitioning is very limited. As in the intro I've explained my goal was to emphasize the different capabilities and levels of abstraction of the different implenetations. If we are strict about what's in the parquet format and ignore all the high level capabilities eg the pyarrow matches about 0% of the format as it can't access, manipulate or create parquet without the intermediate arrow format (unlike in Java and others)

content/en/docs/File Format/implementationstatus.md
alippai: Yes, that's column selection. I think this is the academic language also the Apache Arrow blog refers to that under the same name: https://arrow.apache.org/blog/2022/12/26/querying-parquet-with-millisecond-latency/#projection-pushdown

content/en/docs/File Format/implementationstatus.md
alippai: https://github.com/apache/parquet-format/blob/1603152f8991809e8ad29659dffa224b4284f31b/src/main/thrift/parquet.thrift#L834

content/en/docs/File Format/implementationstatus.md
alippai: https://github.com/apache/parquet-format/blob/1603152f8991809e8ad29659dffa224b4284f31b/src/main/thrift/parquet.thrift#L789

content/en/docs/File Format/implementationstatus.md
alippai: (I don't feel very strongly about this, but also we already have doxygen/sphinx and other API docs which are really good, so... yeah, maybe it's better to have it in a blogpost comparing / demoing the features)

content/en/docs/File Format/implementationstatus.md
alippai: Actually the UUID is good counter example indeed :)

pitrou: We don't support writing UUID from Arrow data indeed. But it probably can be done directly through the Parquet C++ APIs (not PyArrow).

content/en/docs/File Format/implementationstatus.md
pitrou: Hmm, the term "sorting column" is really unfortunate (I find it very confusing: it's not an actual column, just a piece of metadata).

@gszadovszky Thoughts?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
+| Modular encryption                        |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Page CRC32 checksum                       |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| Modular encryption                        |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+
+High level data API-s for parquet feature usage
+===============================================
+
```
pitrou: Need to convert this to Markdown.

content/en/docs/File Format/implementationstatus.md
pitrou: Since this is in a "high level APIs" section I think it's reasonable to include, and informative for the user.

content/en/docs/File Format/implementationstatus.md
pitrou: Might want to add explicit notes to non-obvious items so that the user doesn't have to search for an explanation on the Internet :-)

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
+| Read / Write page metadata and data (2)      |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using projection pushdown       |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using statistics                |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Page pruning using bloom filter              |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+
+* \(1) Ability to construct RowGroup objects from existing RowGroups or raw values (eg. without decoding or decompressing lower level data when read)
```
pitrou: I'm honestly not sure what this means. Reassemble a new Parquet file from existing undecoded column chunks?

content/en/docs/File Format/implementationstatus.md
pitrou: Hmm, why would page pruning use "projection pushdown" but row group pruning use "statistics"? Both actually do projection pushdown, and both use statistics in that goal, am I mistaken?

pitrou: It's a bit confusing though, because it's a different kind of "page pruning" from the below (row-based pruning using statistics or bloom filters).

It's actually not so much "pruning" then "not reading all data at once, just the columns the user is interested in". I assume all reasonable Parquet implementations to do that already, does it need to be mentioned at all?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
++==============================================+=======+========+========+=======+=======+
+| Hive-style partitioning                      |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Partition pruning on the partition column    |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| External column data                         |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| RowGroup Sorting column                      |       |        |        |       |       |
++----------------------------------------------+-------+--------+--------+-------+-------+
+| Read / Write RowGroup metadata and data (1)  |       |        |        |       |       |
```
pitrou: It seems a bit gratuitous to camel-case "RowGroup". Most of this file is English prose, so how about "row group"?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Other format level features
+
++-------------------------------------------+-------+--------+--------+-------+-------+
+|                                           | C++   | Python | Java   | Go    | Rust  |
+|                                           |       |        |        |       |       |
++===========================================+=======+========+========+=======+=======+
+| xxHash Bloom filters                      |       |        |        |       |       |
++-------------------------------------------+-------+--------+--------+-------+-------+
+| bloom filter length                       |       |        |        |       |       |
```
pitrou:
```suggestion
| xxHash-based bloom filters                |       |        |        |       |       |
+-------------------------------------------+-------+--------+--------+-------+-------+
| Bloom filter length                       |       |        |        |       |       |
```

pitrou: Also, not sure what "bloom filter length" is supposed to denote here?

pitrou:
@alippai Perhaps you'd like to start filling in the table for Rust?

alippai:
Thanks @pitrou for the comments. Yes, I’ll address them and fill the table best to my knowledge and online resources

content/en/docs/File Format/implementationstatus.md
alippai: It's an upcoming feature https://github.com/apache/parquet-format/blob/1603152f8991809e8ad29659dffa224b4284f31b/src/main/thrift/parquet.thrift#L766

content/en/docs/File Format/implementationstatus.md
alippai: Yes, that was my intention. AFAIK Both rust and Java solutions support it to some degree

content/en/docs/File Format/implementationstatus.md
alippai: When I read the “pruning” word in DB context I’m always assuming “skipping large chunks”, but I 100% agree that using the same term for both can be confusing

content/en/docs/File Format/implementationstatus.md
alippai: Will do

content/en/docs/File Format/implementationstatus.md
alippai: We can rename it to “integrations” to express this better. Would that be more suitable?

alamb:
I am very interested in helping this post along. Is there any interest in pursuing this PR or would it be better to make a new one? 

>  Thanks @pitrou for the comments. Yes, I’ll address them and fill the table best to my knowledge and online resources

I can certainly fill the one in for Rust

alippai:
I’ll open a new draft later this week

alamb:
FYI https://github.com/apache/parquet-site/pull/53 is a related conversation. Once that PR merges perhaps there will be a more natural location for this chart / location

alamb (commented):
Thank you @alippai 

I have a few suggestions
1. Start the initial PR with a smaller subset of featres (maybe with read/write physical and logical types and compression)
2. Consider distinguishing read and read/write support for the different features (as I think some implementations can read but not create certain encodings or other features)
3. Define terms: define the different implementations (e.g. link to what the "Java" implementation means) and define what  is required to have the implementation in these charts this list. Perhaps use the list on https://parquet.apache.org/docs/overview/

# Smaller Subset

The rationale for a smaller subset is that it will be easier to get consensus. For example,  we won't have to try and precisely describe what "modular encryption support" means 

# "Support"
I think it would also help to define what is required of an implementation to have a "check" in the corresponding feature row.

I think there is an eventual goal to have an automated integration test (e.g. @pitrou 's suggestion in the mailing list https://lists.apache.org/thread/xjg1zrhll5ztqr4n5lv92f6tfwvfb8rm)

However, before that perhaps we could define something less formal, like can read the values of certain files in https://github.com/apache/parquet-testing (e.g. [alltypes_plain.parquet](https://github.com/apache/parquet-testing/blob/master/data/alltypes_plain.parquet)

alippai:
Totally agree, thanks for the guide. What do you think about the non-Apache or other projects? (Duckdb, fastparquet, impala, cudf)

pitrou:
IMHO, any currently maintained open source implementation of Parquet deserves mentioning there. But that also requires involvement from their respective maintainers (we shouldn't expect us Parquet maintainers to make sure the information is up to date).

alamb:
>  Totally agree, thanks for the guide. What do you think about the non-Apache or other projects? (Duckdb, fastparquet, impala, cudf)

Echoing what @pitrou  said, I suggest we add a note somewhere that "this table includes any open source currently maintained implementation of parquet whose maintainers have helped fill it out. If you wish to add a new implementation to this table, please open a PR to do so"

It might also be worth adding a column with some sort of placeholder (`?` for example) for those implementation (as a way of encouraging their help). However, that might be a good thing to do as a follow on PR as well

alippai:
I like both!

emkornfield:
> I think it would also help to define what is required of an implementation to have a "check" in the corresponding feature row.

I worry about this becoming a little bit of a pedantic discussion.  My 2 cents: I think a reasonable approach is to let maintainers decide on checking the box or not.  Maybe we can have a ternary value.  If a maintainer feels it is fully supported it is a check.  If they think reasonable people might be confused it gets '-' or check without footnote to explain the exception.  If it is completely not supported it gets an 'X'.  We can always adjust criteria as we get feedback.

alamb:
> > I think it would also help to define what is required of an implementation to have a "check" in the corresponding feature row.
> 
> I worry about this becoming a little bit of a pedantic discussion. My 2 cents: I think a reasonable approach is to let maintainers decide on checking the box or not. Maybe we can have a ternary value. If a maintainer feels it is fully supported it is a check. If they think reasonable people might be confused it gets '-' or check without footnote to explain the exception. If it is completely not supported it gets an 'X'. We can always adjust criteria as we get feedback.

I agree that it would be best to  start with a relatively lax, low barrier to entry for self reporting 

Over time we can add more stringency  (ideally with automated checks) if/when that would add value

mhaseeb123:
> IMHO, any currently maintained open source implementation of Parquet deserves mentioning there. But that also requires involvement from their respective maintainers (we shouldn't expect us Parquet maintainers to make sure the information is up to date).

Agreed with that. I would be happy to contribute on cudf's behalf!

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| TIME (INT32)                              |       |        |       |       |
+| TIME (INT64)                              |       |        |       |       |
+| TIMESTAMP (INT32)                         |       |        |       |       |
+| TIMESTAMP (INT64)                         |       |        |       |       |
+| INTERVAL                                  |       |        |       |       |
+| JSON                                      |       |        |       |       |
+| BSON                                      |       |        |       |       |
+| LIST                                      |       |        |       |       |
+| MAP                                       |       |        |       |       |
+| UNKNOWN                                   |       |        |       |       |
```
emkornfield: I think this is missing float16?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| Offset index                              |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
+| Page CRC32 checksum                       |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
+
+### High level data API-s for parquet feature usage
+
+| Format                                       | C++   | Java   | Go    | Rust  |
+| -------------------------------------------- | ----- | ------ | ----- | ----- |
+| Hive-style partitioning                      |       |        |       |       |
```
emkornfield: I don't think this is something that is generally part of parquet?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
+---
+title: "Implementation status"
+linkTitle: "Implementation status"
+weight: 8
+---
+### Physical types
```
emkornfield, line 26: it might help to provide guidance on how this should be filled in.  e.g. R for Read Support, W for Write Support and a date or version that it that support started on?

content/en/docs/File Format/implementationstatus.md
alippai: It’s not part of the parquet format, on the other hand dataset and catalog features are part of all the implementations. I’m open to skip it in the first round

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+
+| Data type                                 | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| BOOLEAN                                   |       |        |       |       |
+| INT32                                     |       |        |       |       |
+| INT64                                     |       |        |       |       |
+| INT96                                     |       |        |       |       |
```
etseidl: Should it be noted that INT96 is deprecated?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| PLAIN                                     |       |        |       |       |
+| PLAIN_DICTIONARY                          |       |        |       |       |
+| RLE_DICTIONARY                            |       |        |       |       |
+| RLE                                       |       |        |       |       |
+| BIT_PACKED                                |       |        |       |       |
+| DELTA_BINARY_PACKED                       |       |        |       |       |
+| DELTA_LENGTH_BYTE_ARRAY                   |       |        |       |       |
+| DELTA_BYTE_ARRAY                          |       |        |       |       |
+| BYTE_STREAM_SPLIT                         |       |        |       |       |
```
etseidl, line 78: Should this be split into float/double and int/fixed_len_byte_array, or just use notes if an implementation doesn't yet support the expanded set of data types?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+|                                           | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| xxHash Bloom filters                      |       |        |       |       |
+| bloom filter length                       |       |        |       |       |
+| Statistics min_value, max_value           |       |        |       |       |
+| Column index                              |       |        |       |       |
+| Offset index                              |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
+| Page CRC32 checksum                       |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
```
etseidl, line 102: Add Size Statistics (https://github.com/apache/parquet-format/pull/197)?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Other format level features
+
+|                                           | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| xxHash Bloom filters                      |       |        |       |       |
+| bloom filter length                       |       |        |       |       |
+| Statistics min_value, max_value           |       |        |       |       |
+| Column index                              |       |        |       |       |
+| Offset index                              |       |        |       |       |
```
etseidl: Should these be combined as Page Indexes?

etseidl (commented):
Just a few questions, most about spec changes since this was proposed.

content/en/docs/File Format/implementationstatus.md
mhaseeb123: Alternatively, INT96 can stay as is in the main column as it's a part of the spec and individual implementations can indicate their status?

pitrou: I would rather note "INT96 (deprecated)". It is deprecated in the spec, not only in individual implementations. People not experts in Parquet and looking at this matrix should not get the impression that implementations without INT96 are somehow lagging behind the others.

content/en/docs/File Format/implementationstatus.md
pitrou: +1 on giving guidance. I would support either date or version, though "date of the first release supporting it" sounds better to me (also because at some point I would like to propose that we define calendar-based compatibility profiles).

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
+---
+title: "Implementation status"
+linkTitle: "Implementation status"
+weight: 8
+---
+### Physical types
+
+| Data type                                 | C++   | Java   | Go    | Rust  |
```
pitrou, line 28: I'm not sure that simply naming our implementations "C++", "Java", etc. is very forward-looking, because at some point e.g. DuckDB might want to add their own info here, and they're also written in C++.

Besides, Parquet C++ is also available in Python using PyArrow, in R using R Arrow, and perhaps even in C and Ruby using the GLib bindings.

That said, we can also decide to rename the columns later.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Logical types
+
+| Data type                                 | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| STRING                                    |       |        |       |       |
+| ENUM                                      |       |        |       |       |
+| UUID                                      |       |        |       |       |
+| 8 and 16 bit signed INT                   |       |        |       |       |
+| 8, 16, 32, 64 bit unsigned INT            |       |        |       |       |
```
pitrou: Do these warrant two separate lines?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| DOUBLE                                    |       |        |       |       |
+| BYTE_ARRAY                                |       |        |       |       |
+| FIXED_LEN_BYTE_ARRAY                      |       |        |       |       |
+
+### Logical types
+
+| Data type                                 | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| STRING                                    |       |        |       |       |
+| ENUM                                      |       |        |       |       |
```
pitrou, line 47: I had no idea this even existed :-)

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| DECIMAL (INT32)                           |       |        |       |       |
+| DECIMAL (INT64)                           |       |        |       |       |
+| DECIMAL (BYTE_ARRAY)                      |       |        |       |       |
+| DECIMAL (FIXED_LEN_BYTE_ARRAY)            |       |        |       |       |
+| DATE                                      |       |        |       |       |
+| TIME (INT32)                              |       |        |       |       |
+| TIME (INT64)                              |       |        |       |       |
+| TIMESTAMP (INT32)                         |       |        |       |       |
+| TIMESTAMP (INT64)                         |       |        |       |       |
+| INTERVAL                                  |       |        |       |       |
```
pitrou, line 58: This one doesn't actually exist? (though a number seems to be conventionally reserved for it)

content/en/docs/File Format/implementationstatus.md
pitrou: UNKNOWN doesn't actually exist, does it? It's only used as a placeholder for unknown logical types in some implementations, I think.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| JSON                                      |       |        |       |       |
+| BSON                                      |       |        |       |       |
+| LIST                                      |       |        |       |       |
+| MAP                                       |       |        |       |       |
+| UNKNOWN                                   |       |        |       |       |
+
+### Encoding
```
pitrou:
```suggestion
### Encodings
```

content/en/docs/File Format/implementationstatus.md
pitrou: If we're going to give dates or version numbers as @emkornfield suggested, then this should be split into separate lines.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| DELTA_LENGTH_BYTE_ARRAY                   |       |        |       |       |
+| DELTA_BYTE_ARRAY                          |       |        |       |       |
+| BYTE_STREAM_SPLIT                         |       |        |       |       |
+
+### Compression
+
+| Compression                               | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| UNCOMPRESSED                              |       |        |       |       |
+| SNAPPY                                    |       |        |       |       |
```
pitrou: I would suggest to leave UNCOMPRESSED first, but sort the rest alphabetically for readability. Nobody should care what the physical Thrift numbers are.

content/en/docs/File Format/implementationstatus.md
pitrou, line 102: Why is modular encryption mentioned twice?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| xxHash Bloom filters                      |       |        |       |       |
+| bloom filter length                       |       |        |       |       |
+| Statistics min_value, max_value           |       |        |       |       |
+| Column index                              |       |        |       |       |
+| Offset index                              |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
+| Page CRC32 checksum                       |       |        |       |       |
+| Modular encryption                        |       |        |       |       |
+
+### High level data API-s for parquet feature usage
```
pitrou:
```suggestion
### High level data APIs for Parquet feature usage
```

content/en/docs/File Format/implementationstatus.md
pitrou: Partitioning really seems irrelevant here, especially as it can be applied to other data types such as CSV etc.

Other features such as pruning using statistics, however, are definitely Parquet-related and could be mentioned.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+
+| Format                                       | C++   | Java   | Go    | Rust  |
+| -------------------------------------------- | ----- | ------ | ----- | ----- |
+| Hive-style partitioning                      |       |        |       |       |
+| Partition pruning on the partition column    |       |        |       |       |
+| External column data                         |       |        |       |       |
```
pitrou: What is this?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| RowGroup Sorting column                      |       |        |       |       |
+| Read / Write RowGroup metadata and data (1)  |       |        |       |       |
+| RowGroup pruning using statistics            |       |        |       |       |
+| Read / Write page metadata and data (2)      |       |        |       |       |
+| Page pruning using projection pushdown       |       |        |       |       |
+| Page pruning using statistics                |       |        |       |       |
+| Page pruning using bloom filter              |       |        |       |       |
```
pitrou: Can we try to group related functionality together? Actually, perhaps even make a dedicated table for different kinds of data pruning?

content/en/docs/File Format/implementationstatus.md
emkornfield: It [does](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L401) it is also known as NullType for columns that are all null and therefore the writer couldn't determine the type.

jorisvandenbossche: I have found this UNKNOWN logical type always a confusing name given it's not the same as an unknown logical type ..

content/en/docs/File Format/implementationstatus.md
jorisvandenbossche: While the thrift file indeed indicates it as reserved, it is actually described in the doc: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#interval. Or is that the description for the ConvertedType? (in which case this would be very unclear in the doc)

content/en/docs/File Format/implementationstatus.md
pitrou: Ok, perhaps we can write "UNKNOWN (always null)" to make the table more explanatory?

content/en/docs/File Format/implementationstatus.md
pitrou: Phew. You're right: https://github.com/apache/parquet-format/blob/e91ab5e892391bd179d3c16b7c1cf4fbaeebbfe7/src/main/thrift/parquet.thrift#L399-L400

jorisvandenbossche: On the Parquet C++ side, it seems we also implement the interval LogicalType, but not entirely (it looks like serialization to thrift is not enabled?): https://github.com/apache/arrow/blob/03a960de86d0c3ce8bf299724c94be5b291d85b4/cpp/src/parquet/types.cc#L1364-L1372

etseidl: I did some JIRA diving and found that `INTERVAL` was added by [PARQUET-12](https://issues.apache.org/jira/browse/PARQUET-12) back when converted type and logical type were interchangeable names for the same thing. The separate `LogicalType` struct was added in [PARQUET-906](https://github.com/apache/parquet-format/pull/51) three years later. At that time `INTERVAL` was left out of the new union for [reasons](https://github.com/apache/parquet-format/pull/51#discussion_r118303622)? I'm guessing the description was left in `LogicalTypes.md` to describe the `INTERVAL` converted type. @julienledem can you shed any light here?

jorisvandenbossche: Some other links:

- The open JIRA ticket about adding (other) interval logical types: https://issues.apache.org/jira/browse/PARQUET-675
- A PR that tried to add this but was eventually closed (didn't read it in detail): https://github.com/apache/parquet-format/pull/165
- And the most recent mailing list message about this: https://lists.apache.org/thread/kpt7q57bz180kzqoc2y294wyb1q20x9j

content/en/docs/File Format/implementationstatus.md
emkornfield: Unfortunately, while deprecated some writers like [Spark](https://spark.apache.org/docs/latest/sql-data-sources-parquet.html#configuration) still output by default I believe :(  I agree it should stay here and maybe we can have a subscript detailing this.

content/en/docs/File Format/implementationstatus.md
alamb: Many of the features listed below need to be implemented in the context of a query engine, not just the parquet implementation. 

I personally suggest removing this  section in the first draft to keep things simpler as this page already has lots of great things and we can add this as a follow on PR

Eventually, I think it would make sense to describe features of the reader/writers that enable these features

For example, instead of "row group pruning via statistics"  we might list:
* Writing [`Statitistics`](https://github.com/apache/parquet-format/blob/4227b78c06c23639df8d23c576c0ae92eaf64aff/src/main/thrift/parquet.thrift#L244) for all types
* Reading [`Statitistics`](https://github.com/apache/parquet-format/blob/4227b78c06c23639df8d23c576c0ae92eaf64aff/src/main/thrift/parquet.thrift#L244) for all types
* Specifying reading only certain row groups and skipping others

alamb (approved):
Thank you so much @alippai  and everyone else who has contributed to this PR.


It is my opinion that we should try and address any outstanding comments that are straightforward, but then merge this PR even if there are other more substantial ones unresolved (like guidance https://github.com/apache/parquet-site/pull/34/files#r1635005006). 

We can then iteratively add content over several more PRs and I think some of the comments will be easier to address as we start trying to fill out the chart for implementations. 

I worry that we will run out of ambition if we try to complete this page on the first try (the fact that this PR has been open for more than a year is evidence if telling).

I would be happy to help write a pre-amble and/or try to fill out this table for the Rust implementation, for example, as a follow on PR

This is so exciting to see. I rendered this page locally and it looks good. 
Screenshot 2024-06-14 at 6 08 47 AM

mhaseeb123 (approved):

emkornfield:
> It is my opinion that we should try and address any outstanding comments that are straightforward, but then merge this PR even if there are other more substantial ones unresolved (like guidance https://github.com/apache/parquet-site/pull/34/files#r1635005006).

I agree with this sentiment.  @alippai do you have bandwidth do another pass?  I can merge after another pass (or do so now if you don't have bandwidth for an update).

alippai:
Yes, the input is great and I’ll give it a go tomorrow

xhochy:
Please mark PRs that are ready for review as such. I have not given this as a review as it showed up as Draft in my notification. I will do this now in the next 2h.

content/en/docs/File Format/implementationstatus.md
xhochy: They were all introduced at the same time, and thus I would expect them to be equally supported.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| FIXED_LEN_BYTE_ARRAY                      |       |        |       |       |
+
+### Logical types
+
+| Data type                                 | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| STRING                                    |       |        |       |       |
+| ENUM                                      |       |        |       |       |
+| UUID                                      |       |        |       |       |
+| 8 and 16 bit signed INT                   |       |        |       |       |
```
alkis: This should be 8, 16, 32, 64 (or folded below like @pitrou suggests).

content/en/docs/File Format/implementationstatus.md
alkis: If it is not in the union we might as well call it deprecated and forget it.

Out of curiosity is this type even useful?
- it is 12 bytes which means it will be slower to process than 8 byte types - just like INT96
- given that it stores 3 ints for months, days, millis, it has a range of 357 million years in either direction
- an int64 that stores millis alone has a range of 584 million years

Woot?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Encoding
+
+| Encoding                                  | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| PLAIN                                     |       |        |       |       |
+| PLAIN_DICTIONARY                          |       |        |       |       |
+| RLE_DICTIONARY                            |       |        |       |       |
+| RLE                                       |       |        |       |       |
+| BIT_PACKED                                |       |        |       |       |
```
alkis: `BIT_PACKED (deprecated)`?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+### Compression
+
+| Compression                               | C++   | Java   | Go    | Rust  |
+| ----------------------------------------- | ----- | ------ | ----- | ----- |
+| UNCOMPRESSED                              |       |        |       |       |
+| SNAPPY                                    |       |        |       |       |
+| GZIP                                      |       |        |       |       |
+| LZO                                       |       |        |       |       |
+| BROTLI                                    |       |        |       |       |
+| LZ4                                       |       |        |       |       |
```
alkis: `LZ4 (deprecated)`?

content/en/docs/File Format/implementationstatus.md
alkis: Isn't this supporting reads in other files? https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L868

content/en/docs/File Format/implementationstatus.md
jorisvandenbossche: > we might as well call it deprecated and forget it.

It's not a "deprecated" logical type since it never was a logical type in the first place. It's only a ConvertedType, and my understanding is that all ConvertedTypes are considered as deprecated?

(and this table is listing logical types, not converted types)

So given the above discussion, I think the line should simply be removed.

(it might be a separate question to what extent ConvertedTypes should still be listed on this page)

etseidl: The trouble is, ConvertedType was originally referred to as "Logical Type" (see https://github.com/apache/parquet-format/blame/b3e928e48a013c7170e93a0cce71a776825cb50d/LogicalTypes.md).  The LogicalType construct in the thrift came along much later.

jorisvandenbossche: Yes, but my understanding is that this table specifically lists the elements of the `union LogicalType` in thrift, and not of `enum ConvertedType` (and if that is not the case, that should definitely be clarified)

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,101 @@
...
+| 8, 16, 32, 64 bit unsigned INT            |       |        |       |       |
+| DECIMAL (INT32)                           |       |        |       |       |
+| DECIMAL (INT64)                           |       |        |       |       |
+| DECIMAL (BYTE_ARRAY)                      |       |        |       |       |
+| DECIMAL (FIXED_LEN_BYTE_ARRAY)            |       |        |       |       |
+| DATE                                      |       |        |       |       |
+| TIME (INT32)                              |       |        |       |       |
+| TIME (INT64)                              |       |        |       |       |
+| TIMESTAMP (INT32)                         |       |        |       |       |
+| TIMESTAMP (INT64)                         |       |        |       |       |
```
jorisvandenbossche:
```suggestion
| TIMESTAMP (INT64)                         |       |        |       |       |
```

For TIME we indeed allow both int32/int64 physical type depending on the unit, but for TIMESTAMP I think it is always int64

content/en/docs/File Format/implementationstatus.md
jorisvandenbossche: They are all under the INTEGER logical type, so would indeed list signed and unsigned int as a single line

content/en/docs/File Format/implementationstatus.md
etseidl: My point is, it _is_ a logical type (which is the title of this table), just not in the `LogicalType` union (and as you noted, it's still covered in `LogicalTypes.md`). It's deprecation status is ambiguous. `ConvertedType` is deprecated, and no new types are to be added to it, but readers are still expected to be able to read and understand `ConvertedType`, and writers are still expected to populate it when a mapping exists. All this seven years after deprecation. This is well beyond scope for this PR, but it would be nice if deprecated things actually did disappear from the spec after a certain time (`INT96`, I'm looking at you 👀). Otherwise what's the point of deprecation? Perhaps listing it here, and having few (if any) implementations supporting it would provide the ammunition to kill it off once and for all. 🤷

content/en/docs/File Format/implementationstatus.md
alippai: Removed for now

alippai:
I've learned a ton and I see that many of us saw a few surprises. Looks like the page is needed indeed :)

Following the advice a few of you raised I removed any controversial or too detailed items, we'll handle it in a next PR.

Most of the comments are addressed and if we are happy with the changes, I'll:
1. Rename the existing and add the additional implementations (DuckDB, Impala, Fastparquet etc)
2. Fill them out with initial values and reach out to the communities (if not present here already)
3. Create issues for the followup work (like the read/write distinction, query engine or dataset features, date or version based grouping of the features)

etseidl:
@alippai thanks for updates! I agree with @alamb and @emkornfield that it's the right time to merge this and let the various implementations take a crack at filling in their columns. Questions that arise while doing that can drive the next iteration.

emkornfield:
@alippai this sounds like a good plan.  I think if you are going to fill out some columns it might make sense to do so in this PR to avoid merge conflicts.  It sounds like @alamb and @etseidl want to update for Rust and CUDF respectively but hopefully can coordinate to avoid merge conflicts.

etseidl:
> It sounds like @alamb and @etseidl want to update for Rust and CUDF respectively but hopefully can coordinate to avoid merge conflicts.

Actually @mhaseeb123 will be handling cuDF. I'm just an interested bystander 😅.

mhaseeb123:
> > It sounds like @alamb and @etseidl want to update for Rust and CUDF respectively but hopefully can coordinate to avoid merge conflicts.
> 
> Actually @mhaseeb123 will be handling cuDF. I'm just an interested bystander 😅.

I would be happy to contribute on cuDF's behalf. I am just not fully sure if we are changing column names from language names (C++, Python etc) to implementation names (Arrow, cuDF etc) as @pitrou suggested or filling the rows with implementation names that support a feature in that language?

alippai:
It seems there are no objections. Let me do one more round tomorrow with the update mentioned

content/en/docs/File Format/implementationstatus.md
alamb: I agree -- let's rename the columns later (as we fill out the details) with some name that can be mapped to the implementation (`e.g. `parquet-cpp`, `parquet-java`, etc)

alamb (approved):
Let's get this PR merged!

I started sketching out what a parquet integration suite might look like here: https://github.com/apache/arrow-rs/pull/5956

Once this PR is merged, i will work on honing the integration suite and use that to fill out this table

content/en/docs/File Format/implementationstatus.md
julienledem: Let's add a paragraph at the top with a pointer to each implementation (Java, go, cpp, rust, ...) that will make it easy to add more implementations and clarify which one we're talking about.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -0,0 +1,104 @@
+---
+title: "Implementation status"
+linkTitle: "Implementation status"
+weight: 8
+---
```
julienledem, line 5: Maybe add a legend:
✅ supported
❌ not supported 
[blank] no data 

The main goal being to clarify the difference between missing information and not supported feature.

julienledem (changes requested):

content/en/docs/File Format/implementationstatus.md
alamb: Proposed addition (targeting this PR) in https://github.com/alippai/parquet-site/pull/1

content/en/docs/File Format/implementationstatus.md
alamb: Proposed addition (targeting this PR) in https://github.com/alippai/parquet-site/pull/1

alamb:
I tried to address @julienledem 's comments in  https://github.com/alippai/parquet-site/pull/1

mhaseeb123:
Hi @alamb, thank you for the all work and addressing comments. I saw you added a legend and some description in https://github.com/alippai/parquet-site/pull/1 but I don't see cuDF in there and also am not sure how to exactly fill the table for cuDF. This is confusing especially as cuDF provides C++, Python and (limited? @etseidl) Java Parquet interfaces with GPU-accelerated backends.

Please do note that I am in favor of merging the PR and making things live atm and thinking about this later in a separate PR as well.

emkornfield:
@alippai were you planning on adding anything else here or should we merge?

alamb:
> I saw you added a legend and some description in [alippai#1](https://github.com/alippai/parquet-site/pull/1) but I don't see cuDF in there and also am not sure how to exactly fill the table for cuDF. This is confusing especially as cuDF provides C++, Python and (limited? @etseidl) Java Parquet interfaces with GPU-accelerated backends.

My suggestion is to add a column/columns for cuDF in a future PR when you have information to fill out.

I think the intent of these tables is to describe the parquet implementations (so it lists parquet-cpp, not its python bindings in pyarrow, for example). Perhaps the same model would apply to the parquet reader in cuDF

In terms of how to fill out the table, we are discussing various option for "what does it mean to support a feature" on https://github.com/apache/parquet-format/issues/441

emkornfield:
@alippai thank you for getting this sarted, I'm going to merge now to let others try to fill it out.

alamb:
Here is a fun example of this implementation status work helping drive implementations forward: https://www.linkedin.com/feed/update/urn:li:activity:7415646900902604800/

Screenshot 2026-01-13 at 11 55 40 AM

PR #38, opened 2024-01-19
MINOR: Add link to `LogicalTypes.md`

jasonhorner:
Fixing broken link


content/en/docs/File Format/Types/logicaltypes.md
```diff
@@ -10,4 +10,4 @@ of primitive types to a minimum and reuses parquet's efficient encodings. For
 example, strings are stored as byte arrays (binary) with a UTF8 annotation.
 These annotations define how to further decode and interpret the data.
 Annotations are stored as `LogicalType` fields in the file metadata and are
-documented in LogicalTypes.md.
+documented in [LogicalTypes.md](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md)
```
wgtmac, line 13: This is by purpose. Or at least we should link to a release version: https://github.com/apache/parquet-format/blob/apache-parquet-format-2.10.0/LogicalTypes.md

jasonhorner: is there a way we can do this in a way that will always link to the right version of the docs?  Happy to do the update

wgtmac: We've had a similar discussion since it is painful to update the site for every release: https://lists.apache.org/thread/5qrcz8ps83hnd079dtzd1q6xm31zjbvh

Fokko: I would add the link as is. I don't think the link will break any time soon :)

alamb (approved):
Thank you @jasonhorner @wgtmac and @Fokko 

I took the liberty of merging this PR from production to resolve a conflict

I agree with @Fokko  -- in my opinion having the website properly reflect the latest spec is an improvement over the current state of affairs. 

Figuring out how to properly version documentation for different releases of parquet-format is also important, but seems orthogonal to this PR

I filed another ticket to track:
- https://github.com/apache/parquet-site/issues/126

wgtmac (approved):

wgtmac:
Thanks @alamb for reviving this. Let's move forward.

PR #43, opened 2024-03-05
Dropping go to 1.12 in go.mod

vinooganesh:
This should fix the build. There was an issue that has to do with missing toolchain here: https://github.com/golang/go/issues/62278

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,5 @@
 module github.com/apache/parquet-site
 
-go 1.23
+go 1.12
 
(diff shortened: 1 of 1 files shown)


vinooganesh:
@wgtmac could you +1 and merge this too?

wgtmac (approved):

vinooganesh:
Thanks - looks like this fixed staging: https://parquet.staged.apache.org/!

PR #45, opened 2024-03-06
Force redeploy website

vinooganesh:
@wgtmac - apologies, the last PR didn't contain any core website changes so nothing was redeployed. This should fix that.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "/"
+baseURL = "/" 
 title = "Parquet"
 
 # Language settings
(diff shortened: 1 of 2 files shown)


wgtmac:
Could you also update the asf yaml like https://github.com/apache/parquet-mr/blob/master/.asf.yaml to avoid notifications to dev@ ML? Thanks!

wgtmac (approved):

vinooganesh:
@wgtmac - Ah, so I think the main ones may be coming from git@apache.org, which I believe is a gitbox config. I can look into how to get rid of those notifications, but I think that config are specified outside of this yaml config unfortunately

PR #48, opened 2024-03-08
Fix hugo deprecation warning

deining:
This PR fixes a deprecation warning emitted by hugo:

```
WARN  deprecated: config: languages.en.description: custom params on the language top level
was deprecated in Hugo v0.112.0 and will be removed in a future release. Put the value below
[languages.en.params]. See https://gohugo.io/content-management/multilingual/#changes-in-hugo-01120
```

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -48,12 +48,12 @@ anchor = "smart"
 [languages]
 [languages.en]
 title = "Apache Parquet"
-description = "The Apache Parquet Website"
 languageName ="English"
 contentDir = "content/en"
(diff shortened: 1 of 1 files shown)


vinooganesh:
Hi @deining - this is actually fixed on staging and there will be a pretty large bulk change going to production soon: https://github.com/apache/parquet-site/blob/staging/config.toml#L50. I'd recommend waiting until that change has merged, because many issues have been fixed there.

wgtmac (approved):

PR #56, opened 2024-05-11
Add Dockerfile + instructions on how to preview site using docker rather than installing `hugo` locally

alamb:
In order to make changes to the website and have confidence that we won't break things we should make sure we can see the results of the work locally. 

I can't / don't want to try and figure out how to get a local `hugo` install running locally, and prefer to use docker.

I figured these instructions might help others

BTW I am happy to make a JIRA for this PR, but it isn't clear to me if that is desired or not in Parquet

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 This website is built / powered by [Hugo](https://gohugo.io/), and extended from the [Docsy Theme](https://www.docsy.dev/).
 
-The following steps assume that you have `hugo` installed and working.
+The following steps assume that you have `hugo` installed and working. 
+You can also use docker, see the [Docker section](#docker) for more information.
 
 ## Building and Running Locally
 
@@ -14,7 +15,7 @@ cd parquet-site
 git submodule update --init --recursive
 ```
 
-To build or update your site’s CSS resources, you also need PostCSS to create the final assets. By default npm installs tools under the directory where you run npm install.
+To build or update CSS resources, you also need PostCSS to create the final assets.  By default npm installs tools under the directory where you run npm install.
 
 ```
 npm install -D autoprefixer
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -14,21 +15,61 @@ cd parquet-site
...
+To preview this website site locally, run the following in the root of the directory:
 
 ```shell
 hugo server
 ```
 
-# Release Documentation
+## Building and Running in Docker
+
+If you don't want to install `hugo` and its dependencies local machine, you can use
```
wgtmac:
```suggestion
If you don't want to install `hugo` and its dependencies on local machine, you can use
```

README.md
```diff
@@ -14,21 +15,61 @@ cd parquet-site
...
+docker to preview locally.  First checkout the `parquet-site` explained above 
+and then run:
+
+```shell
+# run docker container mounting the current directory to /parquet-site and exposing port 1313
+docker run -it -v `pwd`:/parquet-site -p 1313:1313  debian:bullseye-slim  bash
```
wgtmac: Is it better to use a dockerfile which is much easier to use? I'm just asking but not required to change. These steps are helpful enough.

wgtmac (approved):
+1

Thanks for the improvement! My only concern is that these steps may be out of sync easily (e.g. when the provided URLs are broken).

README.md
alamb: A docker file is a good idea. I will make one

alamb:
> Thanks for the improvement! My only concern is that these steps may be out of sync easily (e.g. when the provided URLs are broken).

Perhaps we can update the instructions over time if/when they become broken? I am sure there are better ways to make such scripts, but in my opinion this is a step in the right direction

alamb:
Thanks for the review @wgtmac  -- I have implemented your suggestion and created a Dockerfile and updated the instructions to use them.

vinooganesh:
This is a great suggestion and the timing is right. I spend some time a few weeks ago moving the parquet site's docsy dependency to a hugo module, so now they can be managed separately. 

The other thing that we haven't been doing a good job of is maintaining the staging website. I made a bunch of changes to get the `staging` and `production` branch in sync, but staging still isn't heavily used.

alamb:
> The other thing that we haven't been doing a good job of is maintaining the staging website. I made a bunch of changes to get the staging and production branch in sync, but staging still isn't heavily used.

I wonder what the usecase for the staging website is? (maybe we should just not use it?)

FWIW for https://arrow.apache.org/ and https://datafusion.apache.org/ we simply publish to the production version of the site.

Sometimes the staging site might be helpful to host pre-release api docs or something, but I didn't see any on this site 🤔

vinooganesh:
Good question @alamb. Technically the "best practice" from the docsy instructions were to create a staging website so I mostly just followed them when I remade the parquet one. Back then, there was a lot of stuff to work through with hugo builds and migrating from the old jenkins site, so having a place to test was definitely helpful. At this point though, I don't think it's necessary to have the staging site anymore.

wgtmac:
For the staging site, I had a discussion with @gszadovszky here: https://github.com/apache/parquet-site/pull/31#issuecomment-1474023977. I think we can remove the staging site now and use the docker file for debug purpose.

alamb:
Thanks @wgtmac  and @vinooganesh

PR #58, opened 2024-05-12
Remove staging

vinooganesh:
There still needs to be an infra ticket filed to actually delete the `staging` branch (unless a PMC member can delete the branch)

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -18,8 +18,5 @@ github:
   collaborators:  # Note: the number of collaborators is limited to 10
     - vinooganesh
 
-staging:
-  profile: ~
-  whoami:  asf-staging 
 publish:
   whoami:  asf-site


vinooganesh:
cc @wgtmac @gszadovszky @alamb after conversation on https://github.com/apache/parquet-site/pull/56

alamb:
We may also want to update the readme too: https://github.com/apache/parquet-site/blob/production/README.md

vinooganesh:
Yep, there is actually a sequencing of things that need to happen here:
1. Deleting the `asf-staging` brach
2. Deleting the `staging branch` 
3. Deleting the README from the production branch.

The main thing I'm curious about is whether the PMC can delete branches easily from github. If so, it maybe much more straightforward, otherwise will have to file INFRA tickets

alamb:
> The main thing I'm curious about is whether the PMC can delete branches easily from github. If so, it maybe much more straightforward, otherwise will have to file INFRA tickets

In the arrow and datafusion repos, any committer can delete any branch other than the "protected" one (typically the main one)

Thus I suspect someone like @wgtmac  could do so in this repo

wgtmac:
I think I can delete the staging branch. Before that, should we send a notice to the dev ML in case there is any objection? Maybe we can set a deadline and proceed after that. I don't think this worth a formal vote.

BTW, should we do anything to remove the site: https://parquet.staged.apache.org? Or it will be removed automatically after we are done?

alamb:
> I think I can delete the staging branch. Before that, should we send a notice to the dev ML in case there is any objection? Maybe we can set a deadline and proceed after that. I don't think this worth a formal vote.

I agree -- either an email or JIRA ticket is probably good to record the rationale and decision in a more easily discoverable location

> BTW, should we do anything to remove the site: https://parquet.staged.apache.org/? Or it will be removed automatically after we are done?

I suggest we make one final PR to the staging branch to have it push some sort of notice or redirect so https://parquet.staged.apache.org/ redirects to https://parquet.apache.org/

I don't know how to remove it

vinooganesh:
I initially built this to publish to the `output/` dir per https://github.com/apache/infrastructure-asfyaml/tree/main?tab=readme-ov-file#pelican-sub-directories-for-static-output. see: https://github.com/apache/parquet-site/blob/staging/.github/workflows/deploy.yml#L46. We may want to publish an empty commit to the `output` dir of the `asf-staging` branch as just a sanity check cleanup too.

alamb (approved):

wgtmac:
cc @gszadovszky @julienledem

gszadovszky (approved):

wgtmac:
> > I think I can delete the staging branch. Before that, should we send a notice to the dev ML in case there is any objection? Maybe we can set a deadline and proceed after that. I don't think this worth a formal vote.
> 
> I agree -- either an email or JIRA ticket is probably good to record the rationale and decision in a more easily discoverable location
> 
> > BTW, should we do anything to remove the site: https://parquet.staged.apache.org/? Or it will be removed automatically after we are done?
> 
> I suggest we make one final PR to the staging branch to have it push some sort of notice or redirect so https://parquet.staged.apache.org/ redirects to https://parquet.apache.org/
> 
> I don't know how to remove it

Do we want to make additional change, or it is good to merge as is? @vinooganesh @alamb 

I can try to delete staging-related branches once you think fit.

vinooganesh:
I think we can merge as is, if only to see what the behavior will be. It seems pretty low risk, especially since we still have the branches and can revert if needed.

wgtmac:
Sure, let me merge this.

vinooganesh:
Ah okay, so it does seem like we need to clear out the output directory of `asf-staging`: https://parquet.staged.apache.org/

alamb:
> Ah okay, so it does seem like we need to clear out the output directory of `asf-staging`: https://parquet.staged.apache.org/

I suggest we make a PR that directly targets that branch that leaves a pointer to https://parquet.apache.org/ for anyone who stumbles on it

PR #69, opened 2024-07-04
GH-68: Match language from parquet-format after merge of PARQUET-2139

### LINKED ISSUE #68: Update Format section to match format changes from PARQUET-2139
[PARQUET-2139](https://github.com/apache/parquet-format/pull/440) changed the format diagram and some wording that are mirrored [here](https://parquet.apache.org/docs/file-format/). Parquet-site should match those changes.

etseidl:
Closes #68

### DIFF
--- a/content/en/docs/File Format/_index.md
+++ b/content/en/docs/File Format/_index.md
@@ -33,7 +33,7 @@ groups.  The file metadata contains the locations of all the column chunk
 start locations.  More details on what is contained in the metadata can be found
 in the Thrift definition.
 
-File Metadata is written after the data to allow for single pass writing.
+File metadata is written after the data to allow for single pass writing.
 
 Readers are expected to first read the file metadata to find all the column
 chunks they are interested in.  The columns chunks should then be read sequentially.
--- a/content/en/docs/File Format/_index.md
+++ b/content/en/docs/File Format/_index.md
@@ -11,29 +11,29 @@ This file and the thrift definition should be read together to understand the fo
 
 ```
     4-byte magic number "PAR1"
-    <Column 1 Chunk 1 + Column Metadata>
-    <Column 2 Chunk 1 + Column Metadata>
+    <Column 1 Chunk 1>
+    <Column 2 Chunk 1>
     ...
-    <Column N Chunk 1 + Column Metadata>
-    <Column 1 Chunk 2 + Column Metadata>
-    <Column 2 Chunk 2 + Column Metadata>
+    <Column N Chunk 1>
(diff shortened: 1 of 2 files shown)


wgtmac (approved):

content/en/docs/File Format/metadata.md
```diff
@@ -3,8 +3,8 @@ title: "Metadata"
 linkTitle: "Metadata"
 weight: 5
 ---
-There are three types of metadata: file metadata, column (chunk) metadata and page
-header metadata.  All thrift structures are serialized using the TCompactProtocol.
+There are two types of metadata: file metadata, and page header metadata.  All
```
alamb: I recommend providing a link to precisely what these terms are referring to

I think "file metadata" refers to `FileMetadata` https://github.com/apache/parquet-format/blob/ed66e87da9b2d79d6e9262fe37d5eae045c6a639/src/main/thrift/parquet.thrift#L1141

I am not sure what "page header metadata" refers to . Is it `DataPageHeader` https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L580 ? 

If so, maybe we could update this document to use the same terms `FileMetadata` rather than `file metadata` and `DatePageHeader` rather than `page header`

alamb (approved):
Thank you @etseidl  -- I think this is a significant improvement

content/en/docs/File Format/metadata.md
etseidl: It makes more sense when viewed with the image (which has an ERD of the metadata). And this is copied verbatim from the parquet-format `README.md`. But I am in agreement that the parquet-site could provide more information than the format, which is kept terse for a reason. I'll wordsmith this up some.

alamb (approved):
😍 

Thank you @etseidl

content/en/docs/File Format/_index.md
```diff
@@ -11,29 +11,29 @@ This file and the thrift definition should be read together to understand the fo
...
     4-byte magic number "PAR1"
 ```
 In the above example, there are N columns in this table, split into M row
-groups.  The file metadata contains the locations of all the column metadata
+groups.  The file metadata contains the locations of all the column chunk
 start locations.  More details on what is contained in the metadata can be found
 in the Thrift definition.
 
-Metadata is written after the data to allow for single pass writing.
+File Metadata is written after the data to allow for single pass writing.
```
wgtmac:
```suggestion
File metadata is written after the data to allow for single pass writing.
```

PR #90, opened 2024-12-02
Add 1.15.0 release page

Fokko:


wgtmac (approved):

wgtmac:
It is weird that I cannot see any page on 1.15.0 and 1.14.4 from https://parquet.apache.org/blog/

wgtmac:
My bad. They are there but just not sorted by version.

Fokko:
@wgtmac Good catch, it looks like something is off with the date:

image

Let me check!

PR #92, opened 2024-12-04
Run CI on PRs and fail on warnings

Fokko:
The `angolia` search configuration was deprecated, so I took it from the docs:  https://www.docsy.dev/docs/adding-content/search/#algolia-docsearch

```
tml:4:7": execute of template failed: template: 404.html:4:7: executing "404.html" at <partial "head.html" .>: error calling partial: execute of template failed: template: partials/head.html:66:3: executing "algolia/head" at <warnf `Config 'params.algolia_docsearch' is deprecated: use 'params.search.algolia'
      For details, see https://www.docsy.dev/docs/adding-content/search/#algolia-docsearch.`>: error calling warnf: Config 'params.algolia_docsearch' is deprecated: use 'params.search.algolia'
      For details, see https://www.docsy.dev/docs/adding-content/search/#algolia-docsearch.
```

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -3,6 +3,7 @@ name: Build and Deploy Parquet Site
 on:
   push:
     branches: [ production ]
+  pull_request:
 
 jobs:
   Build_and_Deploy_Site:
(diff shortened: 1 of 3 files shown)


wgtmac (approved):
cc @vinooganesh

vinooganesh:
Thanks @Fokko ! You may want to delete that config from here too: https://github.com/apache/parquet-site/blob/e258fa881f2df70ecc62569160e6d8940a93baac/layouts/partials/hooks/body-end.html#L8

Fokko:
@vinooganesh Got it, I wasn't sure if that was required. I copied it to `hugo.toml` to mimic the [example docs](https://docsy.dev/docs/adding-content/search/#configure-algolia-docsearch).  I've removed it from the `html` page.

PR #93, opened 2024-12-04
Add dependabot to update `hugo`

Fokko:
I would recommend merging https://github.com/apache/parquet-site/pull/92 first to make sure that the build passes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,25 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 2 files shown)


vinooganesh:
Looks good to me! @Fokko what are the actual rules for the ASF license headers? Do they need to be on every file?

vinooganesh (approved):

Fokko:
@vinooganesh Thanks for the prompt reply 👍  It is best to put them everywhere. They can be omitted if they are not part of the release (that doesn't make sense in the site context, but I always add them as a habit :). If you run a release, all the files in the `.tar.gz` need to have licenses.

PR #100, opened 2025-02-05
Update matrix for parquet-cpp and parquet-java

wgtmac:


wgtmac:
cc @pitrou @mapleFU @gszadovszky @Fokko @alamb

content/en/docs/File Format/implementationstatus.md
```diff
@@ -45,64 +45,63 @@ Implementations:
...
+| TIME (INT32)                              |  ✅   |  ✅   |       |       |  ✅   |
+| TIME (INT64)                              |  ✅   |  ✅   |       |       |  ✅   |
+| TIMESTAMP (INT64)                         |  ✅   |  ✅   |       |       |  ✅   |
+| INTERVAL                                  |  ✅   |  ✅   |       |       |  ❌   |
+| JSON                                      |  ✅   |  ✅   |       |       |  ❌   |
+| BSON                                      |  ❌   |  ✅   |       |       |  ❌   |
+| LIST                                      |  ✅   |  ✅   |       |       |  ✅   |
+| MAP                                       |  ✅   |  ✅   |       |       |  ✅   |
+| UNKNOWN (always null)                     |  ✅   |  ✅   |       |       |  ✅   |
+| FLOAT16                                   |  ✅   |  ✅   |       |       |  ✅   |
```
gszadovszky: What do we mean by logical types are supported? Is it like we won't throw an exception and return something at least, or we expect to actually represent the related value correctly?
Since `parquet-java` does not have its own representation type system (except the one in the example package that mainly handles the primitive types only), the question is whether at least one binding supports the related type. I am not aware that any binding supports `INTERVAL` or `BSON`. `JSON` is mostly handle as string without any syntax check.

content/en/docs/File Format/implementationstatus.md
```diff
@@ -45,64 +45,63 @@ Implementations:
...
-| LZO                                       |       |       |       |       |  ❌   |
-| SNAPPY                                    |       |       |       |       |  ✅   |
-| ZSTD                                      |       |       |       |       |  ✅   |
+| UNCOMPRESSED                              |  ✅   |  ✅   |       |       |  ✅   |
+| GZIP                                      |  ✅   |  ✅   |       |       |  (R)  |
+| LZ4 (deprecated)                          |  ✅   |  ✅   |       |       |  ❌   |
+| LZ4_RAW                                   |  ✅   |  ✅   |       |       |  ✅   |
+| LZO                                       |  ❌   |  ✅   |       |       |  ❌   |
+| SNAPPY                                    |  ✅   |  ✅   |       |       |  ✅   |
+| ZSTD                                      |  ✅   |  ✅   |       |       |  ✅   |
```
gszadovszky: Why have you removed `BROTLI`?

In the `parquet-java` tests we exclude `LZO` and `LZ4` because they are shipped with Hadoop. Can we say that these are supported?

wgtmac: Oh sorry, I think BROTLI was deleted by accident.

> In the parquet-java tests we exclude LZO and LZ4 because they are shipped with Hadoop. Can we say that these are supported?

I have noticed that as well. I think the feature is supported but by default they are not enabled, right?

content/en/docs/File Format/implementationstatus.md
wgtmac: I don't think won't throw should be considered as supported. Let me remove `INTERVAL` and `BSON`.

Just to confirm that `ENUM` and `UUID` are supported by the Avro binding, right?

mapleFU (approved):
C++ part LGTM

content/en/docs/File Format/implementationstatus.md
```diff
@@ -113,12 +113,12 @@ Implementations:
 
 | Format                                       | C++   | Java  | Go    | Rust  | cuDF  |
 | -------------------------------------------- | ----- | ----- | ----- | ----- | ----- |
-| External column data (1)                     |       |       |       |       |  (W)  |
-| Row group "Sorting column" metadata (2)      |       |       |       |       |  (W)  |
-| Row group pruning using statistics           |       |       |       |       |  ✅   |
-| Row group pruning using bloom filter         |       |       |       |       |  ✅   |
-| Reading select columns only                  |       |       |       |       |  ✅   |
-| Page pruning using statistics                |       |       |       |       |  ❌   |
+| External column data (1)                     |  ❌   |  ✅   |       |       |  (W)  |
```
pitrou: Are we sure that's not supported in C++? I thought Parquet C++ is able to handle pure "metadata" Parquet files that point to other files? Or is this about something else?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -113,12 +113,12 @@ Implementations:
...
+| Row group "Sorting column" metadata (2)      |  ✅   |  ❌   |       |       |  (W)  |
+| Row group pruning using statistics           |  ✅   |  ✅   |       |       |  ✅   |
```
pitrou: Does C++ actually have an API for this? I don't remember.

content/en/docs/File Format/implementationstatus.md
gszadovszky: Yes, `UUID` is mapped to the Avro UUID type, `ENUM` is mapped to string.

But, `JSON` and `FLOAT16` do not seem to be supported either.

content/en/docs/File Format/implementationstatus.md
gszadovszky: Not sure how to represent this or if we want to, but `parquet-java` is capable of handling these compressions (read or write) only if the related hadoop codecs are available in the environment (so not shipped with the library).

content/en/docs/File Format/implementationstatus.md
```diff
@@ -45,64 +45,64 @@ Implementations:
...
+| DATE                                      |  ✅   |  ✅   |       |       |  ✅   |
+| TIME (INT32)                              |  ✅   |  ✅   |       |       |  ✅   |
+| TIME (INT64)                              |  ✅   |  ✅   |       |       |  ✅   |
+| TIMESTAMP (INT64)                         |  ✅   |  ✅   |       |       |  ✅   |
+| INTERVAL                                  |  ✅   |  ❌   |       |       |  ❌   |
+| JSON                                      |  ✅   |  ✅   |       |       |  ❌   |
+| BSON                                      |  ❌   |  ❌   |       |       |  ❌   |
+| LIST                                      |  ✅   |  ✅   |       |       |  ✅   |
+| MAP                                       |  ✅   |  ✅   |       |       |  ✅   |
+| UNKNOWN (always null)                     |  ✅   |  ✅   |       |       |  ✅   |
```
Fokko, line 65: We don't have this one explicitly
```suggestion
| UNKNOWN (always null)                     |  ✅   |  ❌   |       |       |  ✅   |
```

content/en/docs/File Format/implementationstatus.md
wgtmac: I think `FLOAT16` should be considered as supported: https://github.com/apache/parquet-java/pull/1142

content/en/docs/File Format/implementationstatus.md
wgtmac: I have little experience with it. Should we consider it as supported?

content/en/docs/File Format/implementationstatus.md
```diff
@@ -45,64 +45,64 @@ Implementations:
...
+| UUID                                      |  ❌   |  ✅   |       |       |  ❌   |
+| 8, 16, 32, 64 bit signed and unsigned INT |  ✅   |  ✅   |       |       |  ✅   |
+| DECIMAL (INT32)                           |  ✅   |  ✅   |       |       |  ✅   |
+| DECIMAL (INT64)                           |  ✅   |  ✅   |       |       |  ✅   |
+| DECIMAL (BYTE_ARRAY)                      |  ✅   |  ✅   |       |       |  ✅   |
+| DECIMAL (FIXED_LEN_BYTE_ARRAY)            |  ✅   |  ✅   |       |       |  ✅   |
+| DATE                                      |  ✅   |  ✅   |       |       |  ✅   |
+| TIME (INT32)                              |  ✅   |  ✅   |       |       |  ✅   |
+| TIME (INT64)                              |  ✅   |  ✅   |       |       |  ✅   |
+| TIMESTAMP (INT64)                         |  ✅   |  ✅   |       |       |  ✅   |
```
Fokko, line 59: Should we split this out on the unit as well?
```suggestion
| TIMESTAMP (INT64, MICROS)                 |  ✅   |  ✅   |       |       |  ✅   |
```

content/en/docs/File Format/implementationstatus.md
wgtmac: I think the dataset scanner supports this? Or it should not be included as parquet-cpp?

Fokko (commented):
Left some small comments, but this looks great @wgtmac Thanks for updating this 👍

content/en/docs/File Format/implementationstatus.md
pitrou: Ah, yes, the dataset scanner does, but it's another component of Arrow C++, so I'm not sure it qualifies as a Parquet C++ feature.

content/en/docs/File Format/implementationstatus.md
pitrou: Given that it seems to be used in the wild, I would say yes (and we probably have tests for it somewhere, perhaps on the Python side).

content/en/docs/File Format/implementationstatus.md
gszadovszky: So, that's why I asked what supporting a logical type mean in terms of `parquet-java` where we do not have our own representation type system. For `FLOAT16`, we return the containing 2 bytes as a `Binary` in our example package. We do not map this type in any of our bindings.
If we consider `FLOAT16` as supported, then all the others are supported as well.

etseidl: Is there value in adding another option here (`P` perhaps) that means the implementation at least supports the logical type enum value but will return the physical type? As we add logical types to Parquet I think it's useful to know which implementations will at least not error out on more recent additions to the spec.

wgtmac: > If we consider FLOAT16 as supported, then all the others are supported as well.

Yes, that's why I marked all logical types as supported in the initial commit. Developers are able to implement their own bindings to write and read such types by extending the `parquet-java` library.

> Is there value in adding another option here (P perhaps) that means the implementation at least supports the logical type enum value but will return the physical type?

I'm not sure. Maybe it is a minimum requirement for implementation?

content/en/docs/File Format/implementationstatus.md
wgtmac: Let me mark them as unsupported since users cannot use it in the most common cases.

content/en/docs/File Format/implementationstatus.md
wgtmac: That makes sense since it is not in the `libparquet`. Let me mark it as unsupported.

content/en/docs/File Format/implementationstatus.md
wgtmac: Why is that? Any unsupported unit in parquet-java?

content/en/docs/File Format/implementationstatus.md
gszadovszky: @wgtmac, If we do not want to distinguish the support level of e.g. `FLOAT16` and `UUID`, I'm fine having all marked as supported.

wgtmac: OK, fixed.

pitrou: > Yes, that's why I marked all logical types as supported in the initial commit. Developers are able to implement their own bindings to write and read such types by extending the `parquet-java` library.

You should then add a note or asterisk to point out that "supported" means the physical type is returned.

wgtmac: Sounds good. Added asterisk to explain that.

wgtmac:
Do you have any comment with the latest change? @gszadovszky @pitrou @Fokko

gszadovszky (approved):

content/en/docs/File Format/implementationstatus.md
Fokko: No, but I'm not sure if all the other implementations support `nanos` since that was added later on.

Fokko (approved):
Looks good to me, thanks for doing the work @wgtmac

content/en/docs/File Format/implementationstatus.md
wgtmac: It was added almost 7 years ago: https://github.com/apache/parquet-format/commit/b879065ac1bee3fe1d770eb3c4b60ab4267044d7. We can be explicit if we really see an unsupported implementation then.

etseidl (approved):
Thanks @wgtmac! Looks good. I've got rust ready to go once this merges.

wgtmac:
I just merged it. Thanks everyone!

alamb:
> I just merged it. Thanks everyone!

And here is @etseidl 's promised version for parquet-rs:
- https://github.com/apache/parquet-site/pull/101

🥳 🦜

PR #103, opened 2025-02-20
Make format for footnotes more consistent

etseidl:
While filling in the Rust column I noticed that in some places an asterisk is used for table notes, while other places the notes are numerical. This switches the asterisks to `(1)` to allow for additional footnotes should the need arise.


etseidl:
cc @wgtmac

wgtmac (approved):
+1 Thanks!

PR #105, opened 2025-03-02
Update Implementation Status to use Repo Names

platypii:
This PR updates the implementation status table in `content/en/docs/File Format/implementationstatus.md` to use the repository names instead of the language. This is because the repository name is more specific than the language, and this handles the case where there are multiple implementations in the same language.

So for example, instead of `C++`, we now have `arrow`, instead of `Java`, we now have `parquet-java`, and so on. From a practical point of view, most users will probably use parquet through `pyarrow` anyway so maybe this will make it easier for users to understand which libraries support which features. 

I added the language in parentheses after the repository name to help users choose a library in their preferred language.

Note: Stylistically I considered using unicode characters 🇷 and 🇼 to indicate read and write support per @wgtmac's suggestion in #102. But different browsers and operating systems render these characters differently, often different size and style from the existing checkmark (✅) and cross (❌) characters. I decided to stick with the existing characters for consistency.

### DIFF
--- a/content/en/docs/File Format/implementationstatus.md
+++ b/content/en/docs/File Format/implementationstatus.md
@@ -17,96 +17,96 @@ The value in each box means:
 * (blank) no data
 
 Implementations:
-* `C++`: [parquet-cpp](https://github.com/apache/arrow/tree/main/cpp/src/parquet)
-* `Java`: [parquet-java](https://github.com/apache/parquet-java)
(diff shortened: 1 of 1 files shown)


wgtmac:
@alamb WDYT?

alamb:
I think this is a great improvement in clarity -- thank you @platypii  and @wgtmac for the ping

wgtmac (approved):

wgtmac:
I just merged it. Thanks @platypii and @alamb!

PR #107, opened 2025-03-12
Update Implementation Status to note that arrow-rs now supports modular encryption (R)

alamb:
Thanks to @rok , @ggershinsky, @etseidl and @corwinjoy the arrow-rs implementation of Parquet now supports reading from encrypted files 🎉 
- https://github.com/apache/arrow-rs/pull/6637

Let's update the status page

This will be released in about a week as part of the following release, so marking it as a draft until then
- https://github.com/apache/arrow-rs/issues/7107


alamb:
arrow-rs 54.3.0 has been released, so we can/should merge this

wgtmac:
@alamb Thanks for the update! I just merged this.

alamb:
> @alamb Thanks for the update! I just merged this.

Thank you @wgtmac  ❤️

wgtmac:
@zeroshade Do you want to let parquet-go join the party?

PR #115, opened 2025-05-20
Implementation status: hyparquet writer

platypii:
Hyparquet now includes the ability to write parquet files in pure JavaScript. [hyparquet-writer](https://github.com/hyparam/hyparquet-writer)

This PR updates the Implementation Status page to reflect Hyparquet's parquet writing support.

### DIFF
--- a/content/en/docs/File Format/implementationstatus.md
+++ b/content/en/docs/File Format/implementationstatus.md
@@ -118,7 +118,7 @@ Implementations:
 
 | Feature                                   | arrow | parquet-java  | arrow-go | arrow-rs | cudf  | hyparquet | duckdb |
 | ----------------------------------------- | ----- | ------------- | -------- | -------- | ----- | --------- | ------ |
-| External column data (1)                  |  ✅   |  ✅ (*)       |  ❌      |  ❌      |  (W)  |  ✅       |   ❌   |
+| External column data (1)                  |  ✅   |  ✅           |  ❌      |  ❌      |  (W)  |  ✅       |   ❌   |
 | Row group "Sorting column" metadata (2)   |  ✅   |  ❌           |  ✅      |  ✅      |  (W)  |  ❌       |   (R)  |
 | Row group pruning using statistics        |  ❌   |  ✅           |  ✅ (*)  |  ✅      |  ✅   |  ❌       |   ✅   |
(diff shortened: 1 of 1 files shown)


wgtmac:
Thank you for updating it!

wgtmac (approved):

content/en/docs/File Format/implementationstatus.md
```diff
@@ -118,7 +118,7 @@ Implementations:
 
 | Feature                                   | arrow | parquet-java  | arrow-go | arrow-rs | cudf  | hyparquet | duckdb |
 | ----------------------------------------- | ----- | ------------- | -------- | -------- | ----- | --------- | ------ |
-| External column data (1)                  |  ✅   |  ✅           |  ❌      |  ❌      |  (W)  |  ❌       |   ❌   |
+| External column data (1)                  |  ✅   |  ✅ (*)       |  ❌      |  ❌      |  (W)  |  ✅       |   ❌   |
```
wgtmac: Why adding * to parquet-java?

platypii: Sorry that wasn't exactly meant for this PR!

Backstory: I've been implementing support for parquet `file_path` for external column data in hyparquet. And the parquet-java implementation has significant limitations: 1) if any file_path is set, then _every_ column chunk must be external, 2) all file_paths that are external must be the same file, and 3) even with those assumptions I've been unable to make a single example parquet file with external column data that it _can_ read. That might be on me, I'm still investigating.

I can revert if you think so? But it is documented that they have partial support.

platypii: Citation for parquet-java partial `file_path` support: [ParquetMetadataConverter.java#L1814](https://github.com/apache/parquet-java/blob/f3535a6a6/parquet-hadoop/src/main/java/org/apache/parquet/format/converter/ParquetMetadataConverter.java#L1814)

wgtmac: What about clarifying that in a separate PR?

platypii: Reverted the change to parquet-java in this PR. I will open a new PR with the java update later. Thanks for catching that!

PR #120, opened 2025-08-28
Fix release date of format 2.12.0

wgtmac:


wgtmac:
@Fokko Good catch! I made it wrong when calculating the timezone diff :/

PR #144, opened 2025-12-16
Add dates for implementation features.

emkornfield:
This adds dates to the data model for implementation features.

- It then uses those dates to add hyperlinks to implementation notes when available.
- It creates a new table summarizing implementation status per year based on dates.
- Added an entry to extending BYTE_STREAM_SPLIT to additional types.

The data here was backfilled with AI and spot-checked so it is possible some of it might be off.


emkornfield:
Sample table (some data has been adjusted): image

Note the X's have a tool-tip showing which features are missing.

RussellSpitzer:
I like how this looks but I definitely lack the expertise to determine the accuracy of the table as of now.

data/implementations/engines.yaml
```diff
@@ -4,52 +4,195 @@
   language: C++
   repo_url: https://github.com/apache/arrow/tree/main/cpp/src/parquet
   docs_url: https://arrow.apache.org/docs/cpp/parquet.html
-  versions: {}
+  versions:
+    default: "15.0.0"
```
wgtmac: Why default is 15.0.0?

data/implementations/engines.yaml
```diff
@@ -4,52 +4,195 @@
...
 
 - id: arrow-rs
   name: arrow-rs
   display_name: Apache Arrow Rust
   language: Rust
   repo_url: https://github.com/apache/arrow-rs/blob/main/parquet/README.md
   docs_url: null
-  versions: {}
+  versions:
+    default: "15.0.0"
```
wgtmac: Same question here

content/en/docs/File Format/implementationstatus.md
```diff
@@ -12,7 +12,7 @@ of this page by opening an issue or submitting a pull request.
 
 ### Legend
 The value in each box means:
-* ✅: supported. Footnote added when support is partial.
+* ✅: supported. Footnote added when support is partial. When data is available links to release notes are provided on the implementing version.
```
wgtmac:
```suggestion
* ✅: supported. Footnote added when support is partial. When data is available, links to release notes are provided on the implementing version.
```

layouts/shortcodes/implementation-status.html
```diff
@@ -125,11 +240,206 @@ <h3 id="{{ $categoryId }}"><a href="#{{ $categoryId }}">{{ $category.name }}</a>
...
+    <tr>
+      <th>Engine</th>
+      <th>≤2023 Features</th>
+      <th>2024 Features</th>
+      <th>2025 Features</th>
+    </tr>
+  </thead>
+  <tbody>
+    {{- range $engines -}}
+      {{- $engine := . -}}
```
wgtmac, line 328: I found the rendered result for arrow-go is confusing but I find it hard to interpret this logic.

image

emkornfield: hmm, I thought I fixed this let me look into it.

data/implementations/engines.yaml
emkornfield: I thought this was when most of the basics where implemented.  going back it looks like JSON wasn't implemented until 18.0.0 so updated to that.

layouts/shortcodes/implementation-status.html
emkornfield: This was a data issue where 18.4 should have been 2025-07-21.  The code relies on release date for canonical values, and there where some other data issues that I cleaned up (go split off from Arrow mono-repo after 18.0.0 and is only up to 18.5.0).

emkornfield:
@wgtmac thanks for the review, I went back and cleaned up more data after another pass through and generally made the defaults more recent.  I also added a prior Java release and a default so Java support doesn't look as bad in terms of lag.

wgtmac:
Thanks for the update! Generally looks good to me with minor questions as below:

- Is `data/implementations/support/.parquet-java.yaml.swp` added by accident?
- Should we provide a guidance for people to update the implementation status page in the future?

emkornfield:
> Is data/implementations/support/.parquet-java.yaml.swp added by accident?

Yes, need to update the .gitignore for this repo, will do it a separate PR.

> Should we provide a guidance for people to update the implementation status page in the future?

Yes, I think this makes the most sense in the parquet-format CONTRIBUTING guide?  I'll open a separate PR there?

emkornfield:
@wgtmac thanks for review, updated

wgtmac (approved):
Thanks a lot @emkornfield!

emkornfield:
Thanks for the review @wgtmac going to merge.

PR #145, opened 2025-12-19
Rephrase external column data.

emkornfield:
Based on support matrix I believe this was meant to capture Parquet summary files ("_metadata") and not reading external data.

Related format PR: https://github.com/apache/parquet-format/pull/542

### DIFF
--- a/data/implementations/features/api-features.yaml
+++ b/data/implementations/features/api-features.yaml
@@ -1,8 +1,8 @@
 category_id: api-features
 features:
-  - id: api-external-column-data
-    display_name: External column data
-    note: "In parquet.thrift: ColumnChunk->file_path"
+  - id: api-parquet-summary-file
+    display_name: Parquet summary file
+    note: "Files named _metadata that consolidate footers from multiple parquet files."
 
   - id: api-sorting-columns
     display_name: Row group "Sorting column" metadata
--- a/data/implementations/support/hyparquet.yaml
+++ b/data/implementations/support/hyparquet.yaml
@@ -113,8 +113,8 @@ support:
     status: unknown
   format-data-page-v2:
     status: full
-  api-external-column-data:
-    status: full
+  api-parquet-summary-file:
+    status: none
   api-sorting-columns:
     status: none
   api-rowgroup-pruning-stats:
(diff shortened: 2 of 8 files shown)


data/implementations/support/hyparquet.yaml
```diff
@@ -113,8 +113,8 @@ support:
     status: unknown
   format-data-page-v2:
     status: full
-  api-external-column-data:
-    status: full
+  api-parquet-summary-file:
+    status: none
```
emkornfield, line 117: @platypii I looked at hyparquet documentation quickly and didn't see this mentioned but please let me know if I missed it.

data/implementations/features/api-features.yaml
```diff
@@ -1,8 +1,8 @@
 category_id: api-features
 features:
-  - id: api-external-column-data
-    display_name: External column data
-    note: "In parquet.thrift: ColumnChunk->file_path"
+  - id: api-parquet-summary-file
+    display_name: Parquet summary file
+    note: "Files named _metadata that consolidate footers from multiple parquet files."
```
alamb, line 5: I think the reference to the `file_path` field will help keep this specific. So perhaps something like 

```suggestion
    note: "Files named _metadata that consolidate footers from multiple parquet files, stored in ColumnChunk->file_path"
```

alamb (approved):
Thanks @emkornfield 

Assuming https://github.com/apache/parquet-format/pull/542 is correct I think this is a good change too.

I rendered it locally and it looks good to me

Image

data/implementations/features/api-features.yaml
platypii: Is "parquet summary file" really the same as supporting `file_path`?

Maybe summary files are the only known use case of actually using `file_path`? But nothing in the spec says file_path is necessarily associated with "summary files". In fact is there anything in the spec about summary file or is that just an implementation choice some people made?

I'm not opposed to having a row for support of summary files, but that seems like a different question than `file_path` support?

emkornfield: Yeah, I'd err on not including that a reference to file_path here. 

> In fact is there anything in the spec about summary file or is that just an implementation choice some people made?

There is not https://github.com/apache/parquet-format/pull/542 has some more details.

emkornfield:
going to merge.  Thanks for the reviews.

PR #150, opened 2026-01-09
Update Hyparquet features

platypii:
Update Hyparquet supported features matrix. Source: I am the author of hyparquet.

Note: I implemented Variant support yesterday after the parquet sync. Peer pressure works!

image

### DIFF
--- a/data/implementations/engines.yaml
+++ b/data/implementations/engines.yaml
@@ -127,11 +127,20 @@
   repo_url: https://github.com/hyparam/hyparquet
   docs_url: null
   versions:
-    default: "1.23.0"
+    default: "1.0.0"
     releases:
+      "1.0.0":
+        release_date: "2024-06-23"
+        release_notes_url: "https://github.com/hyparam/hyparquet/releases/tag/v1.0.0"
+      "1.19.0":
+        release_date: "2025-09-30"
+        release_notes_url: "https://github.com/hyparam/hyparquet/releases/tag/v1.19.0"
       "1.23.0":
         release_date: "2025-12-10"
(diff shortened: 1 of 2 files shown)


platypii:
(I don't have permission to tag reviewers. @emkornfield @julienledem :pray:)

emkornfield:
Thanks @platypii looks good to me, letting CI run and will merge afterwards.

data/implementations/engines.yaml
```diff
@@ -127,11 +127,20 @@
   repo_url: https://github.com/hyparam/hyparquet
   docs_url: null
   versions:
-    default: "1.23.0"
+    default: "1.0.0"
     releases:
+      "1.0.0":
+        release_date: "2024-06-23"
+        release_notes_url: "https://github.com/hyparam/hyparquet/releases/tag/v1.0.0"
```
emkornfield, line 134: aside: if there is a better URL that actually lists changes, it would be great to include those instead of the generic release links but that is easy enough to modify in a follow-up.

emkornfield (approved):

data/implementations/engines.yaml
platypii: It's all in a single [CHANGELOG.md](https://github.com/hyparam/hyparquet/blob/master/CHANGELOG.md) file. Would you prefer I link to that in every version? Felt a little weird, but happy to do so.

emkornfield: Whatever you prefer, just calling out the links might be useful.

platypii: Ok I'm implementing some more write features, will update the links when I make a future feature update here. Thanks!

PR #152, opened 2026-01-15
Update impl status for bloom filter in arrow-cpp

wgtmac:


pitrou (approved):
+1, thank you @wgtmac !

PR #104, opened 2025-02-24
Add DuckDB to Parquet implementations

szarnyasg:
Heeding the [call](https://x.com/andrewlamb1111/status/1892180295945789900), we at DuckDB Labs (@hannes, @lnkuiper and me) worked on adding the implementation status of Parquet features in DuckDB.

PR #153, opened 2026-01-30
Add Polars to Parquet implementations

coastalwhite:
Similar to #104, we are heeding the [call](https://github.com/pola-rs/polars/issues/26277) to add the implementation status here. Polars has its own implementation that was forked `arrow-rs` many, many years ago, and they have grown apart a lot ever since.

Although I am no longer working at Polars, I was asked by @orlp to come out of retirement (/s) and see if I can fill this in. As I was the person mostly working on Parquet at Polars for the last 1.5 years, I have a decent idea of the status.

If this needs updating, probably @azimafroozeh, @nameexhaustion and @kdn36 are the people to tag.

Referencing:
- pola-rs/polars#22278 to control encoding, compression, page indices
- pola-rs/polars#25185 for the Float16 work
- pola-rs/polars#24320 about the interval types

### DIFF
--- a/data/implementations/engines.yaml
+++ b/data/implementations/engines.yaml
@@ -160,3 +160,16 @@
       "1.4.0":
         release_date: "2025-09-16"
         release_notes_url: "https://github.com/duckdb/duckdb/releases/tag/v1.4.0"
+
+- id: polars
+  name: polars
+  display_name: Polars
+  language: Rust
+  repo_url: https://github.com/pola-rs/polars
+  docs_url: null
+  versions:
+    default: "1.37.0"
+    releases:
+      "1.37.0":
+        release_date: "2026-01-10"
+        release_notes_url: "https://github.com/pola-rs/polars/releases/tag/py-1.37.0"
--- a/data/implementations/support/polars.yaml
+++ b/data/implementations/support/polars.yaml
@@ -1,0 +1,127 @@
+engine_id: polars
+last_updated: "2026-01-30"
+support:
+  physical-boolean:
+    status: full
+  physical-int32:
+    status: full
+  physical-int64:
+    status: full
+  physical-int96:
+    status: read
+  physical-float:
+    status: full
+  physical-double:
+    status: full
+  physical-byte-array:
+    status: full
+  physical-fixed-len-byte-array:
+    status: read
+  logical-string:
+    status: full
+  logical-enum:
+    status: read
+  logical-uuid:
+    status: read
+  logical-int-types:
+    status: full
+  logical-decimal-int32:
+    status: read
+  logical-decimal-int64:
+    status: read
+  logical-decimal-byte-array:
+    status: read
+  logical-decimal-fixed-len-byte-array:
+    status: full
+  logical-float16:
+    status: full
+  logical-date:
+    status: full
+  logical-time-int32:
+    status: read
+  logical-time-int64:
+    status: full
+  logical-timestamp-int64:
+    status: full
+  logical-interval:
+    status: read
+  logical-json:
+    status: none
+  logical-bson:
+    status: none
+  logical-variant:
+    status: none
+  logical-geometry:
+    status: read
+  logical-geography:
+    status: read
+  logical-list:
+    status: full
+  logical-map:
+    status: read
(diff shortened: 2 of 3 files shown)


data/implementations/engines.yaml
```diff
@@ -160,3 +160,16 @@
...
         release_notes_url: "https://github.com/duckdb/duckdb/releases/tag/v1.4.0"
+
+- id: polars
+  name: polars
+  display_name: Polars
+  language: Rust
+  repo_url: https://github.com/pola-rs/polars
+  docs_url: null
+  versions:
+    default: "1.37.0"
```
Fokko, line 171: Should we point to something like _latest_? I see that in the meantime 1.37.1 has been released in the meantime (which is great)!

alamb:
amazing!

data/implementations/engines.yaml
alamb: I think that since this data is mostly used to derive this table, it is ok if it is not complete for all versions

Screenshot 2026-02-04 at 2 37 27 PM

alamb (approved):
Thank you @coastalwhite  (👋 ) and @orlp. It is great to see all the green and (R) 🚀 

I rendered this PR locally and reviewed it for consistency and it looks good to me

Image

I didn't check all the features for accuracy as this page relies on self-reporting and community help to keep it updated.

I'll also post this PR to the parquet mailing list and wait a few more days in case anyone else would like to review it

data/implementations/engines.yaml
emkornfield: @alamb is correct, maybe we should try to rename this field?  "baseline"? (this should be a follow-up)

data/implementations/support/polars.yaml
```diff
@@ -0,0 +1,127 @@
...
+    status: full
+  logical-time-int32:
+    status: read
+  logical-time-int64:
+    status: full
+  logical-timestamp-int64:
+    status: full
+  logical-interval:
+    status: read
+  logical-json:
```
emkornfield, line 48: Just wanted to check for JSON and BSON other engines do support this with the caveat that they just pass through.  Does polars not support them at all?

data/implementations/engines.yaml
emkornfield: @coastalwhite just wanted to verify the X's match your expectation here (2024 I believe specifically had relatively minor changes).  If you render the page, mousing over the X will tell you the missing features.

data/implementations/support/polars.yaml
alamb: I briefly tried to check this, but I couldn't find any relevant files in parquet-testing

I filed a ticket to track adding them
- https://github.com/apache/parquet-testing/issues/101

In the interim, I think we could merge this PR and then adjust the matrix when we get more information

alamb:
I'll plan to merge this PR tomorrow unless others would like time to review

data/implementations/support/polars.yaml
coastalwhite: Looking into it for a second. [Polars indeed should fall through to BinaryView transparently.](https://github.com/pola-rs/polars/blob/3a37c4145ef4f4d46746b08cb5e57757f246ecd2/crates/polars-parquet/src/arrow/read/schema/convert.rs#L164) I have never myself encountered a JSON or BSON file in the wild, so I have never tested it, nor do I remember ever having dealt with any issues about it. (PyArrow, which was my general way of generating Parquet files with specific parameters, also does not support generating them, AFAIK.)

My general rule of thumb is: if it is not structurally tested, it probably doesn't work. I think, therefore, there is a high chance Polars does not handle these correctly at the moment. But I can adjust the matrix to show that the *intent* is that it should work. So that people will file a bug report instead of a feature request.

I am fine merging like this as well, though.

alamb:
Let's merge this one as is and then make other changes as follow ons as necessary. Thank you @coastalwhite and @emkornfield and @Fokko

alamb:
And we are live: https://parquet.apache.org/docs/file-format/implementationstatus/

Screenshot 2026-02-11 at 7 15 14 AM

Issue #146, opened 2026-01-09
[Epic] Blog Posts


alamb:
I am filing a. ticket to track the idea of writing Parquet specific blog posts on features and other general interest topics to the broader Parquet community. 

In my somewhat biased opinion, the main parquet website https://parquet.apache.org/docs  now has good technical reference documentation that will stay up to date with the format (thanks to @emkornfield 's work to synch it with parquet format in https://github.com/apache/parquet-site/issues/140)

However, I think it would be useful to also have content, published on https://parquet.apache.org/blog/
1. aimed at a broader audience to explain Parquet concepts at a slightly higher
2. The intended audience is for those using (or contemplating using) Parquet rather than implementing parquet

 @julienledem also suggested this on the mailing list: https://lists.apache.org/thread/m6p5shlt1hfk1b3h5y1pbqg8tpkwpcgz

> I was thinking that we should talk more about the new features in Parquet.
> Specifically, write blog posts about Variant or Geo types.
> How they came to be, what they enable, how they are integrated in the
> ecosystem.
> The Parquet blog is only the release announcements for now (
> https://parquet.apache.org/blog/) and we can make it an opportunity to
> spotlight some of the recent additions to the project.

Some sources of inspiration:
- https://arrow.apache.org/blog/
- https://openlineage.io/blog
- https://datafusion.apache.org/blog/

Potential topics
- [x] https://github.com/apache/parquet-site/issues/147
- [x] https://github.com/apache/parquet-site/issues/148
- [ ] https://github.com/apache/parquet-site/issues/149
- [ ] #175

devanbenz:
Perhaps one on FSST when that lands in to the spec? 

References: 
https://github.com/apache/arrow-rs/pull/10153
https://github.com/apache/parquet-format/issues/531
https://github.com/apache/arrow-rs/issues/8749

I'd be happy to write it up once the spec is ✔️

PR #154, opened 2026-02-04
Rename "Releases" to "Blog", do not change url

alamb:

As we prepare to add content other than Parquet Java release announcements, I propose we also rename the part of the 

Note that this does not change the existing URL, which is https://parquet.apache.org/blog/

It only changes the text on the home page / nav bar and the page itself as shown in these screen shots

Screenshot 2026-02-04 at 3 12 33 PM

Screenshot 2026-02-04 at 3 12 56 PM

### DIFF
--- a/content/en/blog/_index.md
+++ b/content/en/blog/_index.md
@@ -1,6 +1,6 @@
 ---
-title: "Releases"
-linkTitle: "Releases"
+title: "Blog"
+linkTitle: "Blog"
 menu:
   main:
     weight: 3


alamb:
@Fokko  or @julienledem I wonder if you have a moment to review this PR? I think it is ok, but it would be good to get second set of eyes

julienledem:
That makes sense to me. We could also separate the releases posts to their own page but that can be done later.

wgtmac (approved):

alamb:
Thanks @wgtmac and @julienledem

PR #155, opened 2026-02-04
[BLOG] Template for Geospatial Blog

alamb:
- Related to https://github.com/apache/parquet-site/issues/148
- Related https://github.com/apache/parquet-site/pull/154

Since we don't have any blogs other than release announcements  yet on the parquet site, I wanted to figure out where they would live so that the authors didn't have to figure that out

Thus, this PR is a POC to show how the [Geospatial blog Google Doc](https://docs.google.com/document/d/1JPK0F6Vn4sjXGO4AzrkywOjlj_ybDV_6v0zuaFIIjlk/edit?tab=t.0) from @jiayuasu would work as markdown. 

Hugo made it pretty straightforward. I think the only real question in my mind is what to call the "category" name (I used "feature" but maybe it should be something else? ("other"?)

Here are some annotated screenshots of how this rendered:

Screenshot 2026-02-04 at 3 40 55 PM

Screenshot 2026-02-04 at 3 40 22 PM

PR #156, opened 2026-02-05
[BLOG] Geospatial Blog

### LINKED ISSUE #148: Blog on Geospatial

We recently added Geospatial (Geometry/Geography) types to parquet format -- see https://parquet.apache.org/docs/file-format/types/geospatial/

However, the only documentation that currently exists is the low level technical spec. 

This is somewhat awkward when trying to explain to others what these types are and why they should consider adding Variant support to their systems:  there is no high level overview / link to point people at.

I suggest a blog post on https://parquet.apache.org/blog/  with a high level
1. Explain the usecase of Geospatial types (probably obvious, but highlighting why native parquet support is important, like bounding boxes)
2. Gives a technical overview of the types (e.g. how they are stored)

I would love a story arc that also highlights the fact that this is a major new addition to the Parquet spec and that it has already seen wide adoption https://parquet.apache.org/docs/file-format/implementationstatus/

jiayuasu:
I took @alamb 's [template](https://github.com/apache/parquet-site/pull/155) and created this PR. I hope this is ok.

This idea of this blog post is inspired by [this issue](https://github.com/apache/parquet-site/issues/148) and the initial draft is in this [google doc](https://docs.google.com/document/d/1JPK0F6Vn4sjXGO4AzrkywOjlj_ybDV_6v0zuaFIIjlk/edit?usp=sharing).

- Closes https://github.com/apache/parquet-site/issues/148

Looking forward to having this blog post on Parquet website!


content/en/blog/features/_index.md
```diff
@@ -0,0 +1,6 @@
+
```
alamb: We can probably remove this random blank like (I think that is left over from the template)
```suggestion
```

content/en/blog/features/geospatial.md
```diff
@@ -0,0 +1,152 @@
...
+
+In practice, this means that spatial analytics can finally benefit from the same performance techniques that made Parquet dominant for non-spatial workloads.
+
+![Building Bounding Boxes Visualization](/blog/geospatial/bounding_boxes.png)
+
+**Figure 1:** Visualization of bounding boxes for 130 million buildings stored in a Parquet file from the contiguous U.S. (Microsoft Buildings, file from [geoarrow.org/data](http://geoarrow.org/data), visualization [code](https://gist.github.com/paleolimbot/06303283b42161b57ffc37a8fed60890) here)
+
+## From GeoParquet Metadata to Native Types
+
+Before Parquet adopted GEOMETRY and GEOGRAPHY types in 2025, the [GeoParquet](https://geoparquet.org/) community [1] had already standardized how geometries should be stored in Parquet as early as 2022, using well known binary encoding plus a set of metadata keys. This was an important step because it enabled interoperability across tools.
```
alamb: Given you already have the link to Geoparquet here, I am not sure the `[1]` adds much -- it is also not a hyperlink in the rendered text. I suggest you remove the `[1]` link here and the end of the document and instead just use inline markdown links instead

Image

That being said, I think it is also fine to leave the links here

alamb (commented):
Thank you @jiayuasu  -- I think this looks great!

Here is a preview of what it looks like
Image

If anyone else is interested, here is what the navigation looks like

Screenshot 2026-02-05 at 7 01 32 AM


cc @kylebarron in case you are interested in this content as well

content/en/blog/features/geospatial.md
```diff
@@ -0,0 +1,152 @@
...
+
+Native geospatial types address these issues directly. By making geometry and geography part of the Parquet logical type system, spatial columns become visible to query planners, execution engines, and storage optimizers.
+
+A key benefit is the ability to attach spatial statistics such as bounding boxes to column chunks and row groups. With bounding boxes available in Parquet statistics, engines can skip entire row groups that fall completely outside a query window. This dramatically reduces IO for spatial filters and joins, especially on large datasets.
+
+In practice, this means that spatial analytics can finally benefit from the same performance techniques that made Parquet dominant for non-spatial workloads.
+
+![Building Bounding Boxes Visualization](/blog/geospatial/bounding_boxes.png)
+
+**Figure 1:** Visualization of bounding boxes for 130 million buildings stored in a Parquet file from the contiguous U.S. (Microsoft Buildings, file from [geoarrow.org/data](http://geoarrow.org/data), visualization [code](https://gist.github.com/paleolimbot/06303283b42161b57ffc37a8fed60890) here)
```
RussellSpitzer: Minor note, it's probably better to not link out for images or resources since we can't guarantee they will be available in the future. Just a suggestion though

RussellSpitzer (approved):

content/en/blog/features/geospatial.md
```diff
@@ -0,0 +1,152 @@
...
+This post explains why native geospatial support in Parquet matters and gives a technical overview of how these types are represented and stored.
+
+## Why Geospatial Types Matter in Analytical Storage
+
+Spatial data storage presents unique challenges: a single geometry may represent a point, a road segment, or a complex polygon with thousands of vertices. Queries are also different: instead of simple equality or range filters, users ask spatial questions such as containment, intersection, distance, and proximity in two (XY) or even three (XYZ) dimensions.
+
+Historically, geospatial columns in Parquet were stored as generic binary or string values, with spatial meaning encoded in external metadata. This approach had several limitations.
+
+1. Query engines could not detect a column was GEOMETRY or GEOGRAPHY without an explicit function call by the user (even if the engine supported GEOMETRY or GEOGRAPHY types natively)
+2. Query engines could not apply statistics-based pruning: full Parquet files were required to be read even for spatial queries that returned a small number of rows.
```
2010YOUY01, line 22: It would be great to add a concrete example here to demonstrate spatial pruning, making it more eye-catching and approachable.

I looked at the below bbox visualization code, it's plotting all row groups' bbox from geospatial column statistics in a Parquet file. Perhaps we could build on this by adding a simple example such as:

Query: `SELECT * FROM geometries WHERE ST_INTERSECTS(geo_col, small_bbox)`
(The query window only intersects with a single row group's boundary)

And in the visualization, then highlight the only overlapped row group's bbox, and explain that only the corresponding row group needs to be scanned, demonstrating how spatial pruning dramatically reduces I/O and improves performance.

content/en/blog/features/geospatial.md
```diff
@@ -0,0 +1,152 @@
...
+**Figure 3:** The shortest distance between London and NYC should cross Canada when using the Geography type, whereas the Geometry type incorrectly misses Canada.
+
+Both types integrate into Parquet schemas just like other logical types. From the perspective of a schema definition, a geometry column is no longer an opaque binary field but a typed spatial column.
+
+## How Geospatial Types Are Stored
+
+Although geospatial types are logical constructs, their physical storage follows Parquet's existing columnar design [3]. The following points highlight key aspects of the geospatial type design.
+
+1. **Physical encoding**
+   Geometry and geography values are stored as binary payloads, using Well Known Binary (WKB) encoding. This ensures compatibility across engines and languages.
```
2010YOUY01: We could add a reference link for `WKB`, I think the Wikipedia page has a clear explanation with visualizations:

https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary

content/en/blog/features/geospatial.md
jiayuasu: Thanks. Fixed.

2010YOUY01 (commented):
Thank you, this is a great read! I left a few suggestions for you to consider.

content/en/blog/features/geospatial.md
jiayuasu: Inlined all links. thanks!

content/en/blog/features/geospatial.md
jiayuasu: Thanks. Fixed.

content/en/blog/features/geospatial.md
jiayuasu: Fixed.

paleolimbot (approved):
Thank you @jiayuasu and @alamb for driving this and thanks all for the comments! I took a read through and this is excellent.

steveloughran (commented):
It's been pointed out to me that the coverage matrix doesn't cover statistics/geometry bounding, without which predicate pushdown doesn't work: every rowgroup with the column needs scanning.

https://github.com/apache/parquet-java/blob/7190ab6a571c3cffd31e267042c193e90c0301ad/parquet-column/src/main/java/org/apache/parquet/column/statistics/geospatial/GeospatialStatistics.java#L99

https://github.com/apache/arrow/blob/a82edf90ce66eb9a9a9e3bbac514e5d51f531c1f/cpp/src/parquet/geospatial/util_internal.h#L181

Maybe a "what next?" paragraph

```
Geospatial support in Parquet is still ongoing; as of February 2026 columns statistics collection is incomplete, which means that scanning some types may require reading all the data. Furthermore the query engines themselves need to adopt the new format extensions.

What you do get now is the ability to save geospatial data in Parquet files, with support in those query engines increasing over time.
```

alamb:
> It's been pointed out to me that the coverage matrix doesn't cover statistics/geometry bounding, without which predicate pushdown doesn't work: every rowgroup with the column needs scanning.


 > "Geospatial support in Parquet is still ongoing; as of February 2026 columns statistics collection is incomplete, which means that scanning some types may require reading all the data. Furthermore the query engines themselves need to adopt the new format extensions."

Maybe a more accurate summary is that the column statistics collection is not yet fully integrated into all engines. 

FWIW the Rust Parquet implementation does handle such statistics (thanks to @kylebarron and @paleolimbot  as I recall) -- https://docs.rs/parquet/latest/parquet/format/struct.GeospatialStatistics.html, and I think SedonaDB has already integrated it into their query engine as well. 

Perhaps we can add a line to the https://parquet.apache.org/docs/file-format/implementationstatus/ page for these   (doing so seems to have the effect of pressuring additional ecosystem adoption)

wgtmac (approved):

csringhofer:
Reflecting on the discussion about incomplete statistic support.

I checked a few implementation and while writing statistics for geometries seems to be there in general, I haven't found a single implementation of geography with any edge interpolation algorithm. The rust [implementation](https://github.com/apache/arrow-rs/blob/7dbe58a6e0e18985861db1dfa71507174e838cae/parquet/src/geospatial/accumulator.rs#L151) seems to handle the stats for points (where edge interpolation is not needed) and allows the user to inject its own implementation.

 >Maybe a more accurate summary is that the column statistics collection is not yet fully integrated into all engines.

I agree in case of geometry, but I think that it would make things clearer to mention that for geography this is incomplete, at least in common open source libraries. The blog post mentions "Spatial statistics" as core feature and generally mentions geometry and geography side by side, so the reader may assume that statistics support is widely available for both logical types. This also effect the approach to choosing the best type to use - if bounding boxes are not yet available for geography and per file skipping is critical, then the user should try to build their workload on geometry.

I don't know the status of statistics implementation of geography, but I haven't seen PRs about this, so my assumption is that it may take a significant time to have at least spherical interpolation available widely in Parquet libraries (or extension libraries). I would be happy to be proven wrong :)

Btw the blog was a great read!

steveloughran:
@csringhofer I think @alamb's suggestion about updating the implementation status page might be a tactic, where 
* geometry and geography are somehow separated
* stats collection is a feature to note
* stats usage. Engine and everything which caches aggregate file stats, e.g iceberg v4 manifests

(this'd be so much easier if the flat-earthers were right, though then GPS wouldn't work so measuring locations would be a PITA)

alamb:
Thank you @csringhofer and @steveloughran -- I tried to capture the suggestions on how to improve the status page in a ticket:
- https://github.com/apache/parquet-site/issues/158

> I agree in case of geometry, but I think that it would make things clearer to mention that for geography this is incomplete, at least in common open source libraries. 

I would personally think this would make the page more confusing as
1. I think the point of this blog is to give a high level introduction to the feature from the Parquet perspective. 
2. the implementation status will likely change over time, so such a statement will become outdated

I think a separate blog describing the current state of implementation as of a certain date would be quite valuable for others evaluating potential solutions for their projects

alamb:
Unless there are any objections, I'll plan to update the date and merge this PR (and publish the blog) tomorrow.

alamb:
There seems to be some issue with the parquet site's style sheets / jquery stuff: https://github.com/apache/parquet-site/issues/159

I'll try and find some time to look at this over the next day or two

csringhofer:
>I tried to capture the suggestions on how to improve the status page in a ticket:

Thanks, having a more detailed status page would be a great help for people who try to get an overview / looking for a reference implementation.

>however, I don't think that is inconsistent with the message we are collectively trying to put out with this post (an overview of spatial types in Parquet and celebration of the significant progress we were able to make over the last year).

>the implementation status will likely change over time, so such a statement will become outdated

I see the point - probably there are many things where the implementations could be improved besides geography statistics, and it is not in the scope of the article to go into these.

julienledem (approved):
Thank you for all your contributions, this is going to be a great post!

alamb:
The  https://parquet.apache.org/ site is still kind of broken (at least for me), see
- https://github.com/apache/parquet-site/issues/159

I think we should fix the site before we publish this post and draw more people's attention there

I have a proposed fix here that I would appreciate if someone could help review
- https://github.com/apache/parquet-site/pull/161

content/en/blog/features/geospatial.md
```diff
@@ -0,0 +1,163 @@
...
+
+1. **Physical encoding**
+   Geometry and geography values are stored as binary payloads, using [Well Known Binary (WKB)](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) encoding. This ensures compatibility across engines and languages.
+2. **Spatial statistics**
+   In addition to standard Parquet statistics such as null counts, spatial columns can carry bounding box information. Each row group can record the minimum and maximum extents of the geometries it contains. Query engines can use this information to prune data early when evaluating spatial predicates.
+3. **Engine interoperability**
+   Because the spatial meaning is encoded as a Parquet logical type, engines do not need out of band conventions to interpret the column. A reader that understands Parquet geospatial types can immediately treat the column as a spatial object.
+4. **Coordinate Reference System (CRS) information**
+   CRS information is stored at the file metadata (i.e., type definition) using authoritative identifiers or structured definitions such as EPSG codes or PROJJSON strings.
```
vinooganesh: Nit "CRS information is stored `in` the file metadata"

alamb: Since I was in updating the publish date anyways, I also fixed this typo in fff19958. Hopefully that is ok @jiayuasu  -- if it was a mistake I will make a PR to revert it

alamb:
Thanks to @vinooganesh and @emkornfield I think we are good to publish this blog now 🎉 

I updated the date to Feb 13 and will merge this PR once the CI passes

alamb:
The blog is now live: https://parquet.apache.org/blog/2026/02/13/native-geospatial-types-in-apache-parquet/ 🎉

Issue #159, opened 2026-02-10
Parquet website is visually broken


alamb:
The https://parquet.apache.org/ looks bad:

<img width="1269" height="993" alt="Image" src="https://github.com/user-attachments/assets/4a4054f4-77ce-4789-9809-154deacbbada" />

I looked into it briefly. The page renders correctly locally:

<img width="1472" height="1369" alt="Image" src="https://github.com/user-attachments/assets/c0fdf591-b00b-47da-8f9e-72b2d7f816da" />

But seems to have some sort of cross site loading issue with jquery and the css:

<img width="2269" height="1405" alt="Image" src="https://github.com/user-attachments/assets/018f2396-a474-4345-ab42-3c8298b7b04a" />

I think this was working just great until recently. I am not sure what changed

alamb:
I noticed this first while checking on 
- https://github.com/apache/parquet-site/pull/154

alamb:
It seems like the jquery import is from the docsy template, rather than anything specifically we have in our repo

alamb:
Proposed fix
- https://github.com/apache/parquet-site/pull/161

vinooganesh:
Proposed fix: https://github.com/apache/parquet-site/pull/162

alamb:
Fixed by https://github.com/apache/parquet-site/pull/162

alamb:
To summarize what was wrong:
1. CI was set to automatically update hugo/docsy
2. When https://github.com/apache/parquet-site/pull/154 was merged, and the CI rebuilt the site, a new version of hugo was picked up

This explains why it started happening all of a sudden. Thank you @vinooganesh 

I think it is worth considering some more changes to the site in general, but let's do that as part of follow on issues / without any particular time pressue

alamb:
Filed a ticket to discuss:
- https://github.com/apache/parquet-site/issues/163

PR #160, opened 2026-02-11
Vendor docsy template and add readme explaining how to overrides work

alamb:
- part of https://github.com/apache/parquet-site/issues/159

While debugging https://github.com/apache/parquet-site/issues/159 it was not obvious to me where the overall site tempates were coming from. This may just be my own lack of experience with Hugo/Docsy but as we go and make more extensive modifications to the templates, I think it would help to make it clearer where the source comes from and how to change it

The first commit contains all the files
The second commit contains the change to mod.go and README

Thus:
1. Vendor (aka copy) the docsy theme files 
2. Add instructions in the README.md file that explain how the overrides work


alamb:
@vinooganesh  I would appreciate your feedback on this approach, as you did the original PR

alamb:
I still think we should consider vendoring docsy, but now I see no time pressure for doing so

I created a ticket to discuss the issue
- https://github.com/apache/parquet-site/issues/164

alamb:
superceded by https://github.com/apache/parquet-site/pull/169

PR #161, opened 2026-02-11
Vendor jquery and docsearch to fix website

### LINKED ISSUE #159: Parquet website is visually broken

alamb:
- fixes https://github.com/apache/parquet-site/issues/159
- See also https://github.com/apache/parquet-site/pull/160 for more details about how the site is rendered

I am not sure this is the right / best fix. But it the best I could come up with

# Rationale
Currently the parquet website appears to be broken: https://parquet.apache.org/

image

It works fine when rendered locally

I tracked it down to an issue fetching jquery and docsearch javascript

image

I don't know why this stopped working recently

# Changes
1. Vendor the relevant javascript libraries so they are served directly from this site.
2. I verified locally that after this change the site looks much better

```shell
openssl dgst -sha512 -binary static/vendor/jquery/jquery-3.7.1.min.js | openssl base64 -A
```

Which produces the relevant sha string
```
v2CJ7UaYy4JwqLDIrZUI/4hqeoQieOmAZNXBeQyjo21dadnwR+8ZaIJVT8EE2iyI61OV8e6M8PP2/4hpQINQ/g==
```


layouts/partials/hooks/body-end.html
```diff
@@ -1,10 +0,0 @@
-{{ with .Site.Params.algolia_docsearch }}
```
alamb, line 1: this is already included in the body and head templates, so there is no need to add a second override

layouts/partials/head.html
```diff
@@ -0,0 +1,61 @@
+{{/* cSpell:ignore docsearch opengraph outputformat */ -}}
```
alamb, line 1: This looks like a huge diff, but the diff from the actual template is quite small. I explained how to compute the difference in `README.md`. You can see it here:


```diff
andrewlamb@Andrews-MacBook-Pro-3:~/Software/parquet-site$ diff -du ~/go/pkg/mod/github.com/google/docsy\@v0.12.0/layouts/_partials/head.html layouts/partials/head.html
--- /Users/andrewlamb/go/pkg/mod/github.com/google/docsy@v0.12.0/layouts/_partials/head.html	2026-02-11 07:27:43
+++ layouts/partials/head.html	2026-02-11 09:32:08
@@ -27,9 +27,9 @@
 {{ partial "twitter_cards.html" . -}}
 {{ partialCached "head-css.html" . "head-css-cache-key" -}}
 <script
-  src="https://code.jquery.com/jquery-3.7.1.min.js"
+  src='{{ "vendor/jquery/jquery-3.7.1.min.js" | relURL }}'
   integrity="sha512-v2CJ7UaYy4JwqLDIrZUI/4hqeoQieOmAZNXBeQyjo21dadnwR+8ZaIJVT8EE2iyI61OV8e6M8PP2/4hpQINQ/g=="
-  crossorigin="anonymous"></script>
+></script>
 {{ if .Site.Params.offlineSearch -}}
 <script defer
   src="https://unpkg.com/lunr@2.3.9/lunr.min.js"
@@ -53,9 +53,9 @@
 {{ define "algolia/head" -}}

 {{ if and .Site.Params.search (isset .Site.Params.search "algolia") -}}
-<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/@docsearch/css@3.8.2"
+<link rel="stylesheet" href='{{ "vendor/docsearch/docsearch-3.8.2.css" | relURL }}'
   integrity="sha512-l7pkV1dOURFyHCeH8I4fK9lCkQKuqhlsTCqRl3zktifDlB8oTUJ+mJPgYkK9kHpUut8j1iPquTv32t6hvTPv3g=="
-  crossorigin="anonymous" />
+/>
 {{ end -}}

 {{ if ne .Site.Params.algolia_docsearch nil -}}
```

Note the integrity hash is still the same  (e.g. `integrity="sha512-v2CJ7UaYy4JwqLDIrZUI/4hqeoQieOmAZNXBeQyjo21dadnwR+8ZaIJVT8EE2iyI61OV8e6M8PP2/4hpQINQ/g=="`)

alamb: I have another proposal here to vendor the entire docsy template so it is easier to find / understand
- https://github.com/apache/parquet-site/pull/160

layouts/partials/scripts.html
```diff
@@ -0,0 +1,92 @@
+{{ $needKaTeX  := or .Params.math .Site.Params.katex.enable .Params.chem .Site.Params.chem (.Page.Store.Get "hasKaTeX") (.Page.Store.Get "hasmhchem") -}}
```
alamb, line 1: Likewise here is the actual diff of this file compared to the template base

```diff
andrewlamb@Andrews-MacBook-Pro-3:~/Software/parquet-site$ diff -du ~/go/pkg/mod/github.com/google/docsy\@v0.12.0/layouts/_partials/scripts.html layouts/partials/scripts.html
--- /Users/andrewlamb/go/pkg/mod/github.com/google/docsy@v0.12.0/layouts/_partials/scripts.html	2026-02-11 07:27:43
+++ layouts/partials/scripts.html	2026-02-11 09:32:58
@@ -77,9 +77,8 @@
 {{ partial "hooks/body-end.html" . -}}

 {{ define "algolia/scripts" -}}
-<script src="https://cdn.jsdelivr.net/npm/@docsearch/js@3.8.2"
-  integrity="sha512-lsD+XVzdBI6ZquXc8gqbw0/bgrfIsMJwY/8xvmvbN+U3gZSeG7BXQoCq4zv/yCmntR2GLHtgB+bD4ESPsKIbIA=="
-  crossorigin="anonymous" ></script>
+<script src='{{ "vendor/docsearch/docsearch-3.8.2.min.js" | relURL }}'
+  integrity="sha512-lsD+XVzdBI6ZquXc8gqbw0/bgrfIsMJwY/8xvmvbN+U3gZSeG7BXQoCq4zv/yCmntR2GLHtgB+bD4ESPsKIbIA=="></script>
 <script type="text/javascript">
 const containers = ['#docsearch-0', '#docsearch-1'];
 for (let c of containers) {
```

Again note integrity is the same

README.md
```diff
@@ -77,6 +77,27 @@ hugo server --bind 0.0.0.0 # run the server
 
 You can now preview the site locally on http://localhost:1313/
 
+# Changing the docsy templates
```
alamb, line 80: This explains how to compare the template overrides to the originals

emkornfield:
It would be useful to maybe document why we have vendored these items?  I agree it seems unfortunate but no an expert enough to know if there is a better fix.

alamb:
> It would be useful to maybe document why we have vendored these items? I agree it seems unfortunate but no an expert enough to know if there is a better fix.

Good call -- added in af34cf243472807121e61ea23aed8147bfb652bf

alamb:
@vinooganesh  as the one who setup this site initially, I wonder if you have a brief few moments to look at this PR as well (this one is more urgent than https://github.com/apache/parquet-site/pull/160)

alamb:
Given this is blocking the release of our first content blog and makes the website somewhat unusable, I plan to merge it tomorrow unless there are any objections

If we find a better way I will be happy to revert this and we can use the better approach

vinooganesh:
Taking a look now!

vinooganesh:
@alamb @emkornfield apologies for the delay here, I just put https://github.com/apache/parquet-site/pull/162 up. I'm not sure we need to solve this by vendoring these. This may just be an artifact of the build process. I do owe another docsy upgrade and pass of this to get it back up to date. I was able to repro the issue locally and fix the issue locally, so I think this should work, but want to make sure it actually does on that PR

alamb:
Superceded by 
- https://github.com/apache/parquet-site/pull/162

We can discuss trying to vendor the assets in 
- https://github.com/apache/parquet-site/issues/163

PR #162, opened 2026-02-13
Remove hugo mod get from CI to fix white navbar

vinooganesh:
The navbar on parquet.apache.org has a white background instead of blue.

My speculation is that this happens because hugo mod get in CI upgrades Docsy from v0.12.0 to v0.14.2, which replaces `background: $primary` with a CSS variable that resolves to the body background (white). Since _vendor/ is gitignored, CI doesn't have the vendored Docsy and uses whatever hugo mod get pulls.

Hugo already downloads the pinned versions from go.sum during the build, so hugo mod get isn't needed and was only causing this unwanted upgrade.

Reproed locally:

Broken:
Screenshot 2026-02-12 at 22 38 47

Fixed:
Screenshot 2026-02-12 at 22 39 41

cc @alamb

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -60,9 +60,6 @@ jobs:
 
       - run: npm install
 
-      - name: Download Hugo modules
-        run: hugo mod get
-
       - name: Build
         run: hugo --minify --panicOnWarning
 


emkornfield (approved):
Looks like is worth a try, thanks for the fix.

emkornfield:
For some reason, I'm not sure this triggered a deployment.

emkornfield:
Actually header bar looks like it is back for me.

alamb:
Thank you -- this looks good to me too:

Screenshot 2026-02-13 at 6 30 56 AM

PR #165, opened 2026-02-13
Fix CSP violations blocking jQuery, fonts, and search on production site

### LINKED ISSUE #163: Consider vendoring jquery/css or removing their use from the site
- While debugging https://github.com/apache/parquet-site/issues/159

I would like the local rendered site to match the production site as much as possible so we can test changes before making them live. There are currently several differences

This results in:
1. The fonts don't appear quite the same
2. The local version of the page has search,  but the deployed one does not

I suspect this is related to the fact that there are several resources used when rendering the site locally that are blocked on the production site: https://parquet.apache.org/

I believe the errors are due to the fact that the docsy template we use tries to fetch CSS and jquery from third-party servers  that are not allowed by the apache.org server

Here are the errors:
```
(index):1 Loading the script 'https://code.jquery.com/jquery-3.7.1.min.js' violates the following Content Security Policy directive: "script-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback. The action has been blocked.
(index):1 Loading the stylesheet 'https://fonts.googleapis.com/css?family=Open+Sans:300,300i,400,400i,700,700i&display=swap' violates the following Content Security Policy directive: "style-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback. The action has been blocked.
(index):1 Loading the stylesheet 'https://cdn.jsdelivr.net/npm/@docsearch/css@3.8.2' violates the following Content Security Policy directive: "style-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback. The action has been blocked.
main.min.5ea9bb4146d…cd03b7fa7f2bb4.js:5 Uncaught ReferenceError: jQuery is not defined
    at main.min.5ea9bb4146d…fa7f2bb4.js:5:83140
(index):1 Loading the script 'https://cdn.jsdelivr.net/npm/@docsearch/js@3.8.2' violates the following Content Security Policy directive: "script-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'script-src-elem' was not explicitly set, so 'script-src' is used as a fallback. The action has been blocked.
(index):5 Uncaught ReferenceError: docsearch is not defined
    at (index):5:968
(index):1 Loading the stylesheet 'https://fonts.googleapis.com/css?family=Open+Sans:300,300i,400,400i,700,700i&display=swap' violates the following Content Security Policy directive: "style-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback. The action has been blocked.
(index):1 Loading the stylesheet 'https://cdn.jsdelivr.net/npm/@docsearch/css@3.8.2' violates the following Content Security Policy directive: "style-src 'self' data: blob: 'unsafe-inline' 'unsafe-eval' https://www.apachecon.com/ https://www.communityovercode.org/ https://*.apache.org/ https://apache.org/ https://*.scarf.sh/". Note that 'style-src-elem' was not explicitly set, so 'style-src' is used as a fallback. The action has been blocked.
```

Image

vinooganesh:
Closes #163

The production site at parquet.apache.org is served behind Apache's Content Security Policy, which blocks resources from external CDNs. This causes broken styling, `jQuery is not defined` errors, and completely non-functional search on the live site, even though everything works fine in local dev.

This PR eliminates all external resource dependencies so the site works identically in both environments.

### What was broken

| Blocked resource | Effect on production site |
|---|---|
| `code.jquery.com/jquery-3.7.1.min.js` | `jQuery is not defined` — breaks all JS functionality |
| `fonts.googleapis.com/css?family=Open+Sans` | Font fails to load, falls back unpredictably |
| `cdn.jsdelivr.net/npm/@docsearch/css@3.8.2` | Search styling missing |
| `cdn.jsdelivr.net/npm/@docsearch/js@3.8.2` | `docsearch is not defined` — search completely broken |

### What this PR does

**Vendor jQuery locally** (originally @alamb 's idea, h/t to him) - Downloaded `jquery-3.7.1.min.js` to `static/js/` and created a project-level override of Docsy's `head.html` partial to load it from there instead of the CDN. Same version, just self-hosted.

**Replace Algolia DocSearch with Lunr offline search** - Algolia was doubly broken: blocked by CSP *and* the API keys were removed in a prior commit. Switched to Docsy's built-in Lunr search which is fully client-side with no external dependencies. Also bumped max results to 25 and excerpt length to 200 chars for better usability.

**Disable Google Fonts** - Set `$td-enable-google-fonts: false` so the site uses Bootstrap's system font stack instead of trying to fetch Open Sans from Google. Fonts render immediately with no network request.

### Files changed

- `hugo.toml` - Enable offline search, remove Algolia config
- `assets/scss/_variables_project.scss` - Disable Google Fonts
- `layouts/partials/head.html` (new) - Override Docsy's head to load jQuery and Lunr locally
- `layouts/partials/scripts.html` (new) - Override Docsy's scripts to remove Algolia JS
- `layouts/partials/hooks/head-end.html` - Remove Algolia CSS hook
- `layouts/partials/hooks/body-end.html` - Remove Algolia JS hook
- `static/js/jquery-3.7.1.min.js` (new) - Vendored jQuery
- `static/js/lunr-2.3.9.min.js` (new) - Vendored Lunr

### How I verified

Andrew mentioned he wanted to make sure dev matched prod - I built the site from the clean production branch (before) and confirmed all 4 external CDN references are present. Built again with these changes (after) and confirmed zero external resource loads remain. The dev server serves all JS, CSS, and the search index locally, and search returns results.

Also, no Docsy version changes here, stays on v0.12.0.


layouts/partials/head.html
```diff
@@ -0,0 +1,43 @@
+{{/* cSpell:ignore docsearch opengraph outputformat */ -}}
```
alamb, line 13: Can we add some comments to this file explaining what it does / is used for?

layouts/partials/scripts.html
```diff
@@ -0,0 +1,74 @@
...
+  autoLoader: {
+      manual: true,
+      onReady() {
+        const { autoLoader, builtInPlugins } = window.markmap;
+        autoLoader.transformPlugins = builtInPlugins.filter(plugin => plugin.name !== 'prism');
+      },
+  },
+};
+</script>
+<script src="https://cdn.jsdelivr.net/npm/markmap-autoloader"></script>
```
alamb: Do we also need to vendor this URL?

hugo.toml
```diff
@@ -103,18 +103,15 @@ github_branch = "production"
 # gcs_engine_id = "7e3f91e3eadecceaa"
 
 # Enable Lunr.js offline search
-offlineSearch = false
+offlineSearch = true
```
alamb, line 106: How does the Lunr.js search work? I loaded this page locally and it seems like it still tries to use algolia

Image
Image

assets/scss/_variables_project.scss
```diff
@@ -4,4 +4,7 @@ Add styles or override variables from the theme here.
 
 */
 
-$primary: #56a4ed
+$primary: #56a4ed;
+
+// Disable Google Fonts to comply with apache.org Content Security Policy
```
alamb, line 9: 👍

alamb: Maybe we can create a separate PR just for this change

static/js/jquery-3.7.1.min.js
```diff
@@ -0,0 +1,2 @@
+/*! jQuery v3.7.1 | (c) OpenJS Foundation and other contributors | jquery.org/license */
```
alamb, line 1: I do wonder why we need jquery -- the current site looks ok without it 🤔

alamb (commented):
Thank you for this @vinooganesh 

I am not sure about some of these changes -- maybe we can break them into separate PRs so we can merge the non controversial stuff (like disabling google fonts and removing the body-end and head-end partials)

And then we can have another PR discussion about jquery

vinooganesh:
Good call @alamb ! I'll split these into separate PRs

Issue #164, opened 2026-02-13
Consider vendoring the entire docsy template


alamb:
While debugging https://github.com/apache/parquet-site/issues/159 it was not obvious to me where the overall site templates were coming from. This may just be my own lack of experience with Hugo/Docsy but as we go and make more extensive modifications to the templates, I think it would help to make it clearer where the source comes from and how to change it

Also, I think reducing the amount of third-party content not under (our) source control, that is used to build the site would make it less likely we hit issues such as https://github.com/apache/parquet-site/issues/159 (due to third-party library being updated)

Here is a PR with this this idea
- https://github.com/apache/parquet-site/pull/160

vinooganesh:
Copying my comment here: Hey @alamb - 

Docsy's docs actually recommend going with the hugo go module approach. I'm a little more concerned with vendoring docsy then I am about vendoring lunr and jquery mostly because docsy is a pretty active project. I can imagine upgrading eventually resulting in a huge number of diffs making the CR of actual content more difficult (also I've seen projects that vendor other large repos just decided to never update the dependency because over time it becomes a maintenance pain).

I think the biggest question is how flexible and customizable we want the website to be. I originally chose docsy as just a simple theme that we could pretty easily use/follow that "looked" like other projects (Iceberg was first and I mirrored their setup). If we want a lot more flexibility outside of what docsy provides then I think it's definitely something worth discussing.

PR #170, opened 2026-02-14
Upgrade Docsy theme from v0.12.0 to v0.14.2

vinooganesh:
Upgrades the Docsy theme to v0.14.2 and fixes the visual regressions that came with it:

- Bump Docsy from v0.12.0 to v0.14.2 in go.mod
- Bump Hugo minimum version to 0.155.3 in hugo.toml and deploy.yml
- Fix navbar turning white by setting $td-navbar-bg-color to $primary
- Make navbar transparent on cover pages so it blends with the hero image
- Switch Releases button from btn-secondary to btn-outline-light
- Add navbar_theme = "dark" for proper light text on the blue navbar
- Delete outdated navbar.html override (incompatible with v0.14.2)
- Delete empty hook partials (head-end.html, body-end.html)
- Remove deprecated [blackfriday] config section
- Update head.html and scripts.html overrides to match v0.14.2 (dark mode support, scrollspy patch, updated KaTeX handling) while keeping CSP-compliant local jQuery/Lunr loading

- Related to https://github.com/apache/parquet-site/issues/164


alamb:
@vinooganesh this PR has a "WIP" in its title but is not marked as a draft. Is it ready for review?

alamb:
Also, what do you think about vendoring the theme? As we continue to add overrides to the partials it seems like it would be easier to understand what we have changed 🤔

vinooganesh:
Hey @alamb - it wasn't at the time, but now I think it's ready to go. I'll update the title!

Docsy's docs actually recommend going with the hugo go module approach. I'm a little more concerned with vendoring docsy then I am about vendoring lunr and jquery mostly because docsy is a pretty active project. I can imagine upgrading eventually resulting in a huge number of diffs making the CR of actual content more difficult (also I've seen projects that vendor other large repos just decided to never update the dependency because over time it becomes a maintenance pain). 

I think the biggest question is how flexible and customizable we want the website to be. I originally chose docsy as just a simple theme that we could pretty easily use/follow that "looked" like other projects (Iceberg was first and I mirrored their setup). If we want a lot more flexibility outside of what docsy provides then I think it's definitely something worth discussing.

alamb:
> Docsy's docs actually recommend going with the hugo go module approach. I'm a little more concerned with vendoring docsy then I am about vendoring lunr and jquery mostly because docsy is a pretty active project. 
> I can imagine upgrading eventually resulting in a huge number of diffs making the CR of actual content more difficult (also I've seen projects that vendor other large repos just decided to never update the dependency because over time it becomes a maintenance pain).


My biggest concern is that it will actually be even harder to upgrade to new docsy versions due to our copy/paste/modify overrides of the partial templates. Right now to update docsy and not lose the changes in our overrides we'l have to know how to:
1. Update the docsy version (like this PR)
2. Manually apply any changes to the partials that have been made in the base docsy theme

I am not sure others would even know that we needed to do 2 and so it would be easy to forget

PR #181, opened 2026-05-28
Update hyparquet implementation status

platypii:
Update hyparquet implementation status up to v1.26.0 including:

 - Bloom fliter read/write/pushdown support
 - Variant write support


wgtmac (approved):

wgtmac:
Thank you, @platypii!

PR #183, opened 2026-05-29
Add blog post for ieee754 total order and nan count

### LINKED ISSUE #182: Blog on Floating Point Statistics Improvement (order + Nan counts)
- Related to https://github.com/apache/parquet-format/pull/514
- Related to https://github.com/apache/parquet-format/pull/221
- Related to https://github.com/apache/parquet-format/pull/196

Thanks to the hard work of @JFinis, @wgtmac @etseidl and others we recently updated the Parquet spec to more clearly define how to handle floating point statistics (specifically Nans and ordering of -0.0, +0, etc

I think we should make a post similar to https://parquet.apache.org/blog/2026/02/27/variant-type-in-apache-parquet-for-semi-structured-data/ that describes the backstory of the problem we are solving (in an approachable way) and then the solution that was introduced

Here is one proposed outline
1. Summary of the feature
2. Background on Small Materialized Aggregates / statistics for pruning and parquet including the
3. Background on the wonders of Nans and floating point not having a defined ordering
4. The problems with the old definition for floating point stats (nicely summarized on https://github.com/apache/parquet-format/pull/514)
5. The new solution

In addition to being a nice reference for implementers, I hope  this post to continue the a story arc that highlights news addition we are making to Parquet spec, and and how they are being adopted (e.g. https://parquet.apache.org/docs/file-format/implementationstatus/)

Also it would be a good idea to explain how we arrigved at consensus

wgtmac:
- Closes #182


content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
+---
+title: "Taming Floating-Point Statistics in Apache Parquet: IEEE 754 Total Order and NaN Counts"
+date: 2026-05-29
+description: "How Apache Parquet resolves ambiguous floating-point statistics using IEEE 754 total order and explicit NaN counts for better query performance."
```
alamb:
```suggestion
description: "How the Apache Parquet Community resolved potentially ambiguous floating-point statistics using IEEE 754 total order and explicit NaN counts"
```

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+
+However, floating-point values throw a wrench into this simple model. The IEEE 754 standard defines special values like `NaN` (Not a Number), signed zeros (`-0.0` and `+0.0`), and infinities. Their comparison rules don't play well with the simple "total order" (a strict smaller-to-larger ranking) expected by most data-pruning algorithms. To fix this, the Parquet community recently clarified the standard by combining IEEE 754 total order semantics with an explicit `nan_count` field in the statistics.
+
+The result is a much clearer contract between data writers and readers. Floating-point bounds can now be interpreted consistently, and readers can confidently determine if `NaN` values are present, without having to guess based solely on `min` and `max` bounds.
+
+## Why Floating-Point Statistics Need Special Handling
+
+For integers, strings, and many other straightforward types, Parquet statistics are simple: the writer records the absolute smallest and largest values, and the reader uses those bounds to decide if a query might find a match.
```
alamb, line 17: I think this section would be stronger with a specific example - like maybe two columns of floating points, one with Nans and one without and then a predicate like `where x > 1.0`

My understanding is that the column of floats without the Nan could be  proven to match *all rows* and thus the predicate can be avoiding during execution

However, the column of floats *with* a Nan doesn't match all rows 

Something like this

```
100.0
200.0
Nan.  <-- needs to be filtered out
300.0
```

Previously most parquet writers woudl write stats like
```
min: 100.0
max: 300.0
```

And a clever engine might conclude that *all* rows match (for example the optimization descrbed by @xudong963 in https://datafusion.apache.org/blog/2026/03/20/limit-pruning/) which in this case is incorrect

The engine needs to know if any Nans appear in the data

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+
+Floating-point columns are trickier for two major reasons. First, `-0.0` and `+0.0` are considered equal in normal math operations, yet they possess distinct underlying bit patterns. A data format needs strict rules on how to order these values; otherwise, different libraries might generate conflicting statistics for the exact same underlying data.
+
+Second, `NaN` is completely unordered under standard IEEE 754 comparisons. Expressions like `x < NaN`, `x > NaN`, and `x == NaN` always evaluate to false. If a writer blindly includes `NaN` in ordinary `min` or `max` calculations, the resulting bounds might be useless for skipping data. Conversely, if a writer simply ignores `NaN` values, readers are left in the dark about whether any `NaN`s actually exist in the data block.
```
alamb: As above, I tthink this would be clearer with a motivating example

alamb (commented):
Love it -- I had a suggestion for a specific example that is a little less esoteric than the all nan values example

alamb:
FYI @JFinis and @etseidl  -- perhaps you want to take over this post ( or rewrite it into somehing more in your voice)

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
+---
+title: "Taming Floating-Point Statistics in Apache Parquet: IEEE 754 Total Order and NaN Counts"
+date: 2026-05-29
+description: "How Apache Parquet resolves ambiguous floating-point statistics using IEEE 754 total order and explicit NaN counts for better query performance."
+author: "[Jan Finis](https://github.com/JFinis), [Ed Seidl](https://github.com/etseidl), [Gang Wu](https://github.com/wgtmac)"
```
etseidl: Thanks @wgtmac, but as @alamb knows well, it would be better for all concerned to not include me as an author. Doing so will delay releasing this by months 😮 😭 

Happy to help edit though.

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+
+The result is a much clearer contract between data writers and readers. Floating-point bounds can now be interpreted consistently, and readers can confidently determine if `NaN` values are present, without having to guess based solely on `min` and `max` bounds.
+
+## Why Floating-Point Statistics Need Special Handling
+
+For integers, strings, and many other straightforward types, Parquet statistics are simple: the writer records the absolute smallest and largest values, and the reader uses those bounds to decide if a query might find a match.
+
+Floating-point columns are trickier for two major reasons. First, `-0.0` and `+0.0` are considered equal in normal math operations, yet they possess distinct underlying bit patterns. A data format needs strict rules on how to order these values; otherwise, different libraries might generate conflicting statistics for the exact same underlying data.
+
```
etseidl, line 20:
```suggestion
Parquet's approach to dealing with this ambiguity has been to mandate that a `min` of `0.0` always be written as `-0.0`, and a `max` of `0.0` always be written as `+0.0`, regardless of any sign bits that may be present in the actual data. Readers are advised that `-0.0` may be present even if the `min` is `+0.0`, and `+0.0` may be present even if the max is `-0.0`.

```

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+Second, `NaN` is completely unordered under standard IEEE 754 comparisons. Expressions like `x < NaN`, `x > NaN`, and `x == NaN` always evaluate to false. If a writer blindly includes `NaN` in ordinary `min` or `max` calculations, the resulting bounds might be useless for skipping data. Conversely, if a writer simply ignores `NaN` values, readers are left in the dark about whether any `NaN`s actually exist in the data block.
+
```
etseidl, line 22:
```suggestion
To date Parquet has followed the latter approach, forbidding the inclusion of `NaN` in the statistics. [PR #196](https://github.com/apache/parquet-format/pull/196) provides a detailed overview of the problems inherent in this approach. For instance, consider a page with a max statistic of `0.0` that also contains a `NaN`. A query engine that considers `NaN` to be greater than all values attempts a query with a predicate like `x > 1.0`. If the engine examines the statistics, it will see that the `max` is `0.0`, so it might improperly skip that page, even though it contains at least one row that satisfies the predicate. Without knowledge of the presence or absence of `NaN`, the engine cannot safely perform this type of page pruning for floating point columns.

```

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+These aren't just theoretical edge cases. Query engines rely heavily on these statistics to safely skip large chunks of data. Ambiguous floating-point bounds degrade query performance and can lead to severe inconsistencies. A perfect example of this was highlighted in [PARQUET-2249](https://issues.apache.org/jira/browse/PARQUET-2249), which exposed a critical flaw in how `NaN` values interacted with Parquet's `ColumnIndex` metadata.
```
etseidl:
```suggestion
[PARQUET-2249](https://issues.apache.org/jira/browse/PARQUET-2249), exposed another critical flaw in how `NaN` values interacted with Parquet's `ColumnIndex` metadata.
```

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+
```
etseidl, line 24:
```suggestion
```

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+Ultimately, the community reached a consensus: why not combine the best of both worlds? [PR #514](https://github.com/apache/parquet-format/pull/514), opened in August 2025 and merged on 2026-05-26, successfully merged `IEEE_754_TOTAL_ORDER` with NaN counts. The new order strictly defines how bounds are compared, while `nan_count` clearly flags the presence of `NaN`s. Because no legacy writer uses the new order, the spec now safely *requires* `nan_count` whenever `IEEE_754_TOTAL_ORDER` is used, while gracefully handling older files.
+
+## The Solution
+
+The resulting specification elegantly marries two concepts: an optional `nan_count` field for both `Statistics` and `ColumnIndex`, and the `IEEE_754_TOTAL_ORDER` column order.
+
+`nan_count` records the exact number of `NaN` values within a given scope. Because the field is optional in older files, readers must treat a *missing* `nan_count` differently than a `0`. If missing, readers must cautiously assume `NaN` values *might* be present. If a column is written using `IEEE_754_TOTAL_ORDER`, the writer is forced to provide the `nan_count`.
```
etseidl:
```suggestion
`nan_count` records the exact number of `NaN` values within a given scope. Because the field is optional (or completely missing from older files), readers must treat a *missing* `nan_count` differently than a `0`. If missing, readers must cautiously assume `NaN` values *might* be present. If a column is written using `IEEE_754_TOTAL_ORDER`, the writer is forced to provide the `nan_count`.
```

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+    y_int ^= (((y_int >> 63) as u64) >> 1) as i64;
+    return x_int <= y_int;
+}
+```
+
+The same clever trick applies to 32-bit floats (`f32`): preserve the IEEE bit pattern, flip negative values so their integer representations sort correctly, and compare the resulting integers. The key takeaway for developers is that `min_value` and `max_value` must be computed with the identical total-order semantics advertised in the file's `ColumnOrder`.
+
+Readers, conversely, should treat a missing `nan_count` with caution. Absence doesn't mean zero; it means "unknown," so `NaN` values may lurk inside. When `nan_count` is present, readers can pair it with `min` and `max` bounds to make hyper-efficient, safe pruning decisions.
+
+Implementations that haven't adopted the new rules yet should continue handling older files conservatively, particularly when queries involve `NaN` or signed zeros.
```
etseidl:
```suggestion
Implementations that haven't adopted the new rules yet should continue handling older files conservatively, particularly when queries involve `NaN` or signed zeros, but also in the case of inequalities not involving `NaN`.
```

etseidl (commented):
Just flushing a few suggestions.

I also am a bit uncomfortable with some of the hyperbole 😅 (e.g. "blazing speed", "small but mighty").

wgtmac:
Thank you all for the review! I'll try to address all comments as soon as I can.

wgtmac:
I've accepted all edits. Let me know if there is anything to change. Thanks @alamb and @etseidl!

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+
+Ultimately, the community reached a consensus: why not combine the best of both worlds? [PR #514](https://github.com/apache/parquet-format/pull/514), opened in August 2025 and merged on 2026-05-26, successfully merged `IEEE_754_TOTAL_ORDER` with NaN counts. The new order strictly defines how bounds are compared, while `nan_count` clearly flags the presence of `NaN`s. Because no legacy writer uses the new order, the spec now safely *requires* `nan_count` whenever `IEEE_754_TOTAL_ORDER` is used, while gracefully handling older files.
```
Fokko:
```suggestion
Ultimately, the community reached a consensus: why not combine the best of both worlds? [PR #514](https://github.com/apache/parquet-format/pull/514), opened in August 2025 and merged in May 2026, successfully merged `IEEE_754_TOTAL_ORDER` with `NaN` counts. The new order strictly defines how bounds are compared, while `nan_count` clearly flags the presence of `NaN`s. Because no legacy writer uses the new order, the spec now safely *requires* `nan_count` whenever `IEEE_754_TOTAL_ORDER` is used, while gracefully handling older files.
```

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+
+Readers, conversely, should treat a missing `nan_count` with caution. Absence doesn't mean zero; it means "unknown," so `NaN` values may lurk inside. When `nan_count` is present, readers can pair it with `min` and `max` bounds to make hyper-efficient, safe pruning decisions.
+
+Implementations that haven't adopted the new rules yet should continue handling older files conservatively, particularly when queries involve `NaN` or signed zeros, but also in the case of inequalities not involving `NaN`.
+
+## Conclusion
+
+By embracing IEEE 754 total order and `nan_count`, Apache Parquet now boasts a much clearer, robust foundation for floating-point statistics. This update preserves the blazing speed of predicate pushdown while finally taming the edge cases: `NaN` values are accurately counted, signed zeros have their rightful place, and all floating-point bit patterns can be ordered deterministically.
+
+It’s a small but mighty refinement to the Parquet format. It boosts interoperability across different programming languages and query engines, giving readers the precise information they need to prune data without ever having to guess.
```
Fokko: I would turn this around:
```suggestion
It’s a small but mighty refinement to the Parquet format. It boosts interoperability across different programming languages and query engines, giving readers the precise information they need to prune data with confidence.
```
Otherwise it looks like we were guessing before :)

Fokko (approved):
Two small comments, but apart from that this looks great! Thanks @wgtmac for writing the blog post, I think it would be awesome to get this out to the broader audience.

etseidl (approved):
Thanks @wgtmac, this looks great!

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+
+## The Solution
+
+The resulting specification elegantly marries two concepts: an optional `nan_count` field for both `Statistics` and `ColumnIndex`, and the `IEEE_754_TOTAL_ORDER` column order.
+
+`nan_count` records the exact number of `NaN` values within a given scope. Because the field is optional (or completely missing from older files), readers must treat a *missing* `nan_count` differently than a `0`. If missing, readers must cautiously assume `NaN` values *might* be present. If a column is written using `IEEE_754_TOTAL_ORDER`, the writer is forced to provide the `nan_count`.
+
+This new total order applies exclusively to physical `FLOAT`, `DOUBLE`, or logical `FLOAT16` columns. It defines a strict, deterministic ordering for bit patterns with these key properties:
+
+1. `-0.0` is ordered strictly before `+0.0`.
```
mapleFU, line 45: I'm not an expert, in legacy IEEE 754, `-0.0 == +0.0`, and filter could works here. Would it matter when a new filter causing this to be non-equal? Would the writer or filter caller take more care on this?

content/en/blog/features/ieee754-order.md
```diff
@@ -0,0 +1,88 @@
...
+    let mut y_int = y.to_bits() as i64;
+    x_int ^= (((x_int >> 63) as u64) >> 1) as i64;
+    y_int ^= (((y_int >> 63) as u64) >> 1) as i64;
+    return x_int <= y_int;
+}
+```
+
+The same clever trick applies to 32-bit floats (`f32`): preserve the IEEE bit pattern, flip negative values so their integer representations sort correctly, and compare the resulting integers. The key takeaway for developers is that `min_value` and `max_value` must be computed with the identical total-order semantics advertised in the file's `ColumnOrder`.
+
+Readers, conversely, should treat a missing `nan_count` with caution. Absence doesn't mean zero; it means "unknown," so `NaN` values may lurk inside. When `nan_count` is present, readers can pair it with `min` and `max` bounds to make hyper-efficient, safe pruning decisions.
```
mapleFU, line 72: `It means "unknown",` ?

content/en/blog/features/ieee754-order.md
wgtmac: That's a good question. So the implication to implementations is that special care should be taken when a literal in a predicate is 0. cc @etseidl

content/en/blog/features/ieee754-order.md
wgtmac: I think `"unknown,"` is more preferable (in American English).

content/en/blog/features/ieee754-order.md
mapleFU: Maybe it's not a problem in some implementation when converting the type to arrow min-max statistics. But when using parquet comparator to do column filtering, this might need change filter to `or(eq(-0.0), eq(+0.0))`?

wgtmac: IMO, parquet comparator should just keep its current behavior because that's what IEEE754 total order is specified. High-level filter processing should know what does total order mean and deal with that as it doesn't have ambiguity as before.

content/en/blog/features/ieee754-order.md
mapleFU: Oh, learned from it now!

mapleFU (approved):

content/en/blog/features/ieee754-order.md
etseidl: > So the implication to implementations is that special care should be taken when a literal in a predicate is 0.

I think how engines treat -0 will dictate how they interpret the statistics. But at least now the statistics are well defined. In the past, a min of `0.0` was represented as `-0.0`, which would give incorrect information to engines that distinguish between positive and negative zero.

wgtmac:
Thanks all for the review! Let me merge this.

alamb:
The site is live: https://parquet.apache.org/blog/2026/05/29/taming-floating-point-statistics-in-apache-parquet-ieee-754-total-order-and-nan-counts/

PR #186, opened 2026-06-04
Document Parquet Features by Version

alamb:
## Preview

Rendered preview: https://alamb.github.io/parquet-site/docs/file-format/versions/

## Rationale
- part of https://github.com/apache/parquet-format/issues/384
- See the related [mailing list thread](https://lists.apache.org/thread/st8l40z5n4cx5c22rcog50ws4pkzdc3s).

Adding new backwards-incompatible changes to the Parquet specification requires a way to communicate which systems support which features. Parquet already has a v1/v2 versioning scheme, but it is poorly documented and has caused significant confusion across the ecosystem. Clearly explaining the current (imperfect) scheme is valuable on its own and might also help a potential future V3 rollout.

## Changes

Add a versions page to parquet.apache.org explaining:
1. The current versioning scheme and how Parquet versions relate to each other
2. The version/date each feature was added to the spec
3. Links to relevant releases and mailing list discussions

## Screenshots

Screenshot 2026-06-10 at 3 51 43 PM
Screenshot 2026-06-10 at 3 51 52 PM
Screenshot 2026-06-10 at 3 52 02 PM

### DIFF
--- a/content/en/docs/File Format/versions.md
+++ b/content/en/docs/File Format/versions.md
@@ -31,9 +31,6 @@ suffer, but the reader does not fail. Examples:
 **Forward incompatible** features make the data **unreadable** to older software.
 Examples:
 
-Backwards incompatible features make the data **unreadable** to older software
-that does not support them. Examples:
-
 * **New encodings** (e.g. the `DELTA_*` encodings, `BYTE_STREAM_SPLIT`,
   `RLE_DICTIONARY`): a reader that does not implement them cannot decode the
   column values.
@@ -115,9 +112,10 @@ Forward incompatible features and the format version each became available in:
 | [Modular encryption] |  | ✅ | [2.7.0] | [2.6.0..2.7.0] | [Approved 2019-01-16] |
 
 
-> **Note:** Compression codecs and modular encryption are not gated by the
-> `version` field — a reader needs support for the specific codec or for
-> encryption regardless of version.
+> **Note:** Files with an [encrypted footer] use different magic bytes (`PARE`
+> instead of `PAR1`), making it clear to readers they must support modular
+> encryption to read the file; [plaintext footer] files use `PAR1` so legacy
+> readers can still read their unencrypted columns.
 
 ## Forward compatible additions
 
@@ -169,6 +167,8 @@ new information.
 [DELTA_BYTE_ARRAY]: https://github.com/apache/parquet-format/blob/master/Encodings.md#delta-strings-delta_byte_array--7
 [BYTE_STREAM_SPLIT]: https://github.com/apache/parquet-format/blob/master/Encodings.md#byte-stream-split-byte_stream_split--9
 [Modular encryption]: https://github.com/apache/parquet-format/blob/master/Encryption.md
+[encrypted footer]: https://github.com/apache/parquet-format/blob/master/Encryption.md#54-encrypted-footer-mode
+[plaintext footer]: https://github.com/apache/parquet-format/blob/master/Encryption.md#55-plaintext-footer-mode
 [xxHash-based bloom filters]: https://github.com/apache/parquet-format/blob/master/BloomFilter.md
 [Page index]: https://github.com/apache/parquet-format/blob/master/PageIndex.md
 [FLOAT16]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#float16
--- a/content/en/docs/File Format/versions.md
+++ b/content/en/docs/File Format/versions.md
@@ -133,27 +133,27 @@ new information.
 | [Binary protocol extensions] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2024-09-06]                                     |
 | [IEEE 754 total order and NaN counts] | not yet released | [#514] | [Approved 2026-05-26]                                     |
 | [LogicalType union] | [2.4.0] | [2.3.1..2.4.0] | Supersedes `ConvertedType` enum<br/>deprecated in [2.9.0] |
-| [STRING] | [1.0.0] | [1.0.0][tree-1.0.0] |                                                           |
-| [ENUM] | [2.0.0] | [1.0.0..2.0.0] |                                                           |
-| [UUID] | [2.6.0] | [2.5.0..2.6.0] |                                                           |
-| [Signed and unsigned integer logical types] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
+| [STRING (BYTE_ARRAY)] | [1.0.0] | [1.0.0][tree-1.0.0] |                                                           |
+| [ENUM (BYTE_ARRAY)] | [2.0.0] | [1.0.0..2.0.0] |                                                           |
+| [UUID (FIXED_LEN_BYTE_ARRAY(16))] | [2.6.0] | [2.5.0..2.6.0] |                                                           |
+| [Signed and unsigned integer logical types (INT32, INT64)] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
 | [DECIMAL (INT32)] | [2.1.0] | [2.0.0..2.1.0] |                                                           |
 | [DECIMAL (INT64)] | [2.1.0] | [2.0.0..2.1.0] |                                                           |
 | [DECIMAL (BYTE_ARRAY)] | [2.1.0] | [2.0.0..2.1.0] |                                                           |
 | [DECIMAL (FIXED_LEN_BYTE_ARRAY)] | [2.1.0] | [2.0.0..2.1.0] |                                                           |
-| [FLOAT16] | [2.10.0] | [2.9.0..2.10.0] | [Approved 2023-10-13]                                     |
-| [DATE] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
+| [FLOAT16 (FIXED_LEN_BYTE_ARRAY(2))] | [2.10.0] | [2.9.0..2.10.0] | [Approved 2023-10-13]                                     |
+| [DATE (INT32)] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
 | [TIME (INT32)] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
 | [TIME (INT64)] | [2.4.0] | [2.3.1..2.4.0] |                                                           |
 | [TIMESTAMP (INT64)] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
 | [Nanosecond TIME/TIMESTAMP] | [2.6.0] | [2.5.0..2.6.0] |                                                           |
-| [INTERVAL] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
-| [JSON] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
-| [BSON] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
+| [INTERVAL (FIXED_LEN_BYTE_ARRAY(12))] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
+| [JSON (BYTE_ARRAY)] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
+| [BSON (BYTE_ARRAY)] | [2.2.0] | [2.1.0..2.2.0] |                                                           |
 | [VARIANT] | [2.12.0] | [2.11.0..2.12.0] | [Approved 2025-08-24]                                     |
 | [Variant shredding] | [2.12.0] | [2.11.0..2.12.0] | [Approved 2025-08-24]                                     |
-| [GEOMETRY] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09]                                     |
-| [GEOGRAPHY] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09]                                     |
+| [GEOMETRY (BYTE_ARRAY)] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09]                                     |
+| [GEOGRAPHY (BYTE_ARRAY)] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09]                                     |
 | [LIST] | [1.0.0] | [1.0.0][tree-1.0.0] |                                                           |
 | [MAP] | [1.0.0] | [1.0.0][tree-1.0.0] |                                                           |
 | [UNKNOWN (always null)] | [2.4.0] | [2.3.1..2.4.0] |                                                           |
@@ -171,10 +171,10 @@ new information.
 [plaintext footer]: https://github.com/apache/parquet-format/blob/master/Encryption.md#55-plaintext-footer-mode
 [xxHash-based bloom filters]: https://github.com/apache/parquet-format/blob/master/BloomFilter.md
 [Page index]: https://github.com/apache/parquet-format/blob/master/PageIndex.md
-[FLOAT16]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#float16
+[FLOAT16 (FIXED_LEN_BYTE_ARRAY(2))]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#float16
 [VARIANT]: https://github.com/apache/parquet-format/blob/master/VariantEncoding.md
-[GEOMETRY]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry
-[GEOGRAPHY]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geography
+[GEOMETRY (BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry
+[GEOGRAPHY (BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geography
 
 [1.0.0]: https://github.com/apache/parquet-format/releases/tag/parquet-format-1.0.0
 [2.0.0]: https://github.com/apache/parquet-format/releases/tag/parquet-format-2.0.0
@@ -223,28 +223,26 @@ new information.
 [Bloom filter length]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L933
 [Page CRC32 checksum]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L829
 [Size statistics]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L202
-[IEEE 754 Column Order]: https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L1061
 [LogicalType union]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L471
 [Nanosecond TIME/TIMESTAMP]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L352
-[STRING]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#string
-[ENUM]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#enum
-[UUID]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#uuid
-[8, 16, 32, 64 bit signed and unsigned INT]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#signed-integers
+[STRING (BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#string
+[ENUM (BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#enum
+[UUID (FIXED_LEN_BYTE_ARRAY(16))]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#uuid
 [DECIMAL (INT32)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
 [DECIMAL (INT64)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
 [DECIMAL (BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
 [DECIMAL (FIXED_LEN_BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
-[DATE]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#date
+[DATE (INT32)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#date
 [TIME (INT32)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#time
 [TIME (INT64)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#time
 [TIMESTAMP (INT64)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#timestamp
-[INTERVAL]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#interval
-[JSON]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#json
-[BSON]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#bson
+[INTERVAL (FIXED_LEN_BYTE_ARRAY(12))]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#interval
+[JSON (BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#json
+[BSON (BYTE_ARRAY)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#bson
 [LIST]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists
 [MAP]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps
 [UNKNOWN (always null)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#unknown-always-null
-[Signed and unsigned integer logical types]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#signed-integers
+[Signed and unsigned integer logical types (INT32, INT64)]: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#signed-integers
 [IEEE 754 total order and NaN counts]: https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L1061
 
 [BOOLEAN]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L33
--- a/content/en/docs/File Format/versions.md
+++ b/content/en/docs/File Format/versions.md
@@ -1,6 +1,6 @@
 ---
 title: "Parquet format versions"
-linkTitle: "Format versions"
+linkTitle: "Features and Versions"
 weight: 9
 ---
 
@@ -52,13 +52,12 @@ issues][closing-out-2.0].
 ## `parquet-format` release versions
 
 The Thrift definition is released independently of implementations such as
-parquet-java or arrow-rs, following the Apache release process and
-[semantic versioning]:
+parquet-java or arrow-rs, following the Apache release process. Note that
+release numbering **DOES NOT FOLLOW** [semantic versioning]:
 
 1. The major version corresponds to the [`thrift FileMetadata`] `version` field.
-2. Minor releases (e.g. `2.10.0` to `2.11.0`) may add compatible
-   features, but never incompatible ones. The minor version is not recorded in the
-   file itself.
+2. Minor releases (e.g. `2.10.0` to `2.11.0`) sometimes contain forward incompatible 
+   features. The minor version is not recorded in the file itself.
 
 ## Adding new features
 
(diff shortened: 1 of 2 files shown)


content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+[FLOAT]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L37
+[DOUBLE]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L38
+[BYTE_ARRAY]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L39
+[FIXED_LEN_BYTE_ARRAY]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L40
+[Data Page V1]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L671
+[Data Page V2]: https://github.com/apache/parquet-format/blob/96656a543a2165d57cc1c9abefaad7f9aeb563a5/src/main/thrift/parquet.thrift#L724
+[BIT_PACKED (deprecated)]: https://github.com/apache/parquet-format/blob/master/Encodings.md#bit-packed-deprecated-bit_packed--4
+[BYTE_STREAM_SPLIT<br/>(Additional Types)]: https://github.com/apache/parquet-format/blob/master/Encodings.md#byte-stream-split-byte_stream_split--9
```
Fokko, line 253: Was the `<br/>` intentional?

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
+---
+title: "Parquet format versions"
+linkTitle: "Versions"
```
Fokko:
```suggestion
linkTitle: "Format versions"
```

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+| [GZIP] | ✅ | ✅ | [1.0.0] | [1.0.0][tree-1.0.0] |  |
+| [LZO] | ✅ | ✅ | [1.0.0] | [1.0.0][tree-1.0.0] |  |
+| [BROTLI] |  | ✅ | [2.4.0] | [2.3.1..2.4.0] |  |
+| [LZ4 (deprecated)] |  | ✅ | [2.4.0] | [2.3.1..2.4.0] |  |
+| [LZ4_RAW] |  | ✅ | [2.9.0] | [2.8.0..2.9.0] |  |
+| [ZSTD] |  | ✅ | [2.4.0] | [2.3.1..2.4.0] |  |
+| [Modular encryption] |  | ✅ | [2.7.0] | [2.6.0..2.7.0] | [Approved 2019-01-16] |
+
+
+> **Note:** Compression codecs and modular encryption are not gated by the
```
Fokko: Encryption has a different magic string `PARE` instead of `PAR1`. Should we mention that here?

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+
+[parquet dev mailing list]: https://lists.apache.org/list.html?dev@parquet.apache.org
+[semantic versioning]: https://semver.org/
+[`thrift FileMetadata`]: https://github.com/apache/parquet-format/blob/c42c2cb4ecfccb38153375e24b702a82fd763cc0/src/main/thrift/parquet.thrift#L1365-L1373
+[here]: https://github.com/apache/parquet-format/blob/master/CONTRIBUTING.md#additionschanges-to-the-format
+[closing-out-2.0]: https://lists.apache.org/thread/0bdyyb7qobrxx94x8v7t5z7g2ksnpyr2
+
+## Backwards incompatible features by version
+
+Backwards incompatible features and the format version each became available in:
```
etseidl: Per https://github.com/apache/parquet-format/blob/master/CONTRIBUTING.md#compatibility-and-feature-enablement these are forward incompatible changes (old reader can't read files produced by new writers). I don't know if there have been any changes that are not backwards compatible (new reader can't read files from old writers).

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+| [LZ4_RAW] |  | ✅ | [2.9.0] | [2.8.0..2.9.0] |  |
+| [ZSTD] |  | ✅ | [2.4.0] | [2.3.1..2.4.0] |  |
+| [Modular encryption] |  | ✅ | [2.7.0] | [2.6.0..2.7.0] | [Approved 2019-01-16] |
+
+
+> **Note:** Compression codecs and modular encryption are not gated by the
+> `version` field — a reader needs support for the specific codec or for
+> encryption regardless of version.
+
+## Backwards compatible additions
```
etseidl: Same here...Forwards compatible

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+| [TIME (INT64)] | [2.4.0] | [2.3.1..2.4.0] |  |
+| [TIMESTAMP (INT64)] | [2.2.0] | [2.1.0..2.2.0] |  |
+| [INTERVAL] | [2.2.0] | [2.1.0..2.2.0] |  |
+| [JSON] | [2.2.0] | [2.1.0..2.2.0] |  |
+| [BSON] | [2.2.0] | [2.1.0..2.2.0] |  |
+| [VARIANT] | [2.12.0] | [2.11.0..2.12.0] | [Approved 2025-08-24] |
+| [Variant shredding] | [2.12.0] | [2.11.0..2.12.0] | [Approved 2025-08-24] |
+| [GEOMETRY] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09] |
+| [GEOGRAPHY] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09] |
+| [LIST] | [1.0.0] | [1.0.0][tree-1.0.0] |  |
```
etseidl: My notes say LIST and MAP were not formally codified with the three-level structure until 2.3.1 (https://github.com/apache/parquet-format/commit/0e2e0a469f3b2dd6b53210a89c851cbbf663fd6f)

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+> `version` field — a reader needs support for the specific codec or for
+> encryption regardless of version.
+
+## Backwards compatible additions
+
+Files written with these features can be read by all readers, but older
+readers may not understand the new information.
+
+| Feature | Released in | Source | Notes |
+| ------------------------------------------- | ----------------------------- | --- | ------------------------- |
```
etseidl: Also from my own notes:
- Addition of the LogicalType union in favor of ConvertedType enum was in 2.4.0.
- Formal deprecation of ConvertedType in 2.9.0
- Addition of NANOS to TimeUnit 2.6.0

etseidl (commented):
Flushing some comments. Thanks for taking this on @alamb, I think this is an important addition.

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+[closing-out-2.0]: https://lists.apache.org/thread/0bdyyb7qobrxx94x8v7t5z7g2ksnpyr2
+
+## Backwards incompatible features by version
+
+Backwards incompatible features and the format version each became available in:
+
+* **V1**: the original Parquet format (1.0).
+* **V2**: format version 2.0.
+
+| Feature | V1 | V2 | Released in | Source | Notes |
```
emkornfield: could we somehow integrate this as YAML data + rendering like we do for feature compatibility?

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+| ------------------------------------------- | ----------------------------- | --- | ------------------------- |
+| [xxHash-based bloom filters] | [2.7.0] | [2.6.0..2.7.0] | [Approved 2019-09-09] |
+| [Bloom filter length] | [2.10.0] | [2.9.0..2.10.0] |  |
+| [Page index] | [2.4.0] | [2.3.1..2.4.0] |  |
+| [Page CRC32 checksum] | [1.0.0] | [1.0.0][tree-1.0.0] |  |
+| [Size statistics] | [2.10.0] | [2.9.0..2.10.0] | [Approved 2023-11-14] |
+| [Geospatial statistics] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09] |
+| [Binary protocol extensions] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2024-09-06] |
+| [IEEE 754 total order and NaN counts] | not yet released | [#514] | [Approved 2026-05-26] |
+| [STRING] | [1.0.0] | [1.0.0][tree-1.0.0] |  |
```
Fokko: We mention the physical types below, should we do it here as well?
```suggestion
| [STRING (BYTE_ARRAY)] | [1.0.0] | [1.0.0][tree-1.0.0] |  |
```

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,258 @@
...
+| [ENUM] | [2.0.0] | [1.0.0..2.0.0] |  |
+| [UUID] | [2.6.0] | [2.5.0..2.6.0] |  |
```
Fokko: We mention the physical types below, should we do it here as well?
```suggestion
| [UUID (FIXED_LEN_BYTE_ARRAY(16))] | [2.6.0] | [2.5.0..2.6.0] |  |
```

Fokko (approved):
Comments are mostly bikeshedding, but I think this would be something great to add. Thanks for taking the time to put this together 🙌

content/en/docs/File Format/versions.md
alamb: Yes -- due to my whitespace obsession. 

I was obsessing about how the table rendered and I needed to add a manual line break to avoid an overly wide first column

content/en/docs/File Format/versions.md
alamb: I am happy to convert this to be data driven rendering rather than an explicit table (rather I would  get Claude to do it). 

But here it may make less sense as I expect this page to be the only consumer/producer of this data, and the features don't seem to need the same type of cross referencing / forced deduplication of the status page

content/en/docs/File Format/versions.md
alamb: Thank you -- great catch. Will update the content

content/en/docs/File Format/versions.md
alamb: From what I can tell, LIST and MAP are actually defined in the 1.0.0 spec (using the old, deprecated "ConvertedType" annotations)
* `LIST`: https://github.com/apache/parquet-format/blob/parquet-format-1.0.0/src/thrift/parquet.thrift#L57-L59
* `MAP`: https://github.com/apache/parquet-format/blob/parquet-format-1.0.0/src/thrift/parquet.thrift#L51-L52

It seems like https://github.com/apache/parquet-format/commit/0e2e0a469f3b2dd6b53210a89c851cbbf663fd6f added the types to the LogicalType.md type structure 

🤔

content/en/docs/File Format/versions.md
alamb: In 7f37c8e2. I will admit I found the "forward compatibility" more confusing, but I think consistency is more important

content/en/docs/File Format/versions.md
alamb: Also in https://github.com/apache/parquet-site/commit/7f37c8e25352ee3282b70ee7d149a33911495e5c

content/en/docs/File Format/versions.md
etseidl: Yes, the original LIST was underspecified, and implementers diverged in how they were represented in the schema. To enable nullable lists with nullable elements, the current 3-level structure in the schema was introduced. We still suffer with having to parse older, now non-compliant lists (see for example https://github.com/apache/arrow-rs/pull/8496).

I'm just pointing this out for historical purposes 🤓

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,265 @@
...
+* **Logical type annotations** such as `VARIANT`: an older reader reads the
+  underlying physical column (e.g. `BYTE_ARRAY`) as raw bytes without applying
+  the logical type.
+
+**Forward incompatible** features make the data **unreadable** to older software.
+Examples:
+
+Backwards incompatible features make the data **unreadable** to older software
+that does not support them. Examples:
+
```
etseidl:
```suggestion
```

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,265 @@
...
+  `RLE_DICTIONARY`): a reader that does not implement them cannot decode the
+  column values.
+* **Data Page V2 headers**: a reader that only understands `DataPageHeader`
+  cannot parse `DataPageHeaderV2` pages.
+
+[classifies changes]: https://github.com/apache/parquet-format/blob/master/CONTRIBUTING.md#compatibility-and-feature-enablement
+
+## `FileMetadata` version field
+
+Each Parquet file has a `version` field in the [`thrift FileMetadata`] that
```
etseidl: I've said this elsewhere, but I don't think we can rely on a version in the metadata, as it cannot convey changes to the metadata itself.

I think we should just say this field conveys no meaningful information any longer.

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,265 @@
...
+declares which features the file may use, and thus what a reader **must** support
+to read it.
+
+**Note**: Many writers set the version field to `1` even for files that use
+format 2.0 features, which has caused [confusion and interoperability
+issues][closing-out-2.0].
+
+## `parquet-format` release versions
+
+The Thrift definition is released independently of implementations such as
```
etseidl, line 51: This is not currently true (as the table below demonstrates). Beyond the bump to major 2, we've only bumped minor, whether the changes contained are breaking or not. This does need to change IMO.

etseidl (commented):
A few more nits

content/en/docs/File Format/versions.md
alamb: > Also from my own notes:
> 
> * Addition of the LogicalType union in favor of ConvertedType enum was in 2.4.0.
> * Formal deprecation of ConvertedType in 2.9.0
> * Addition of NANOS to TimeUnit 2.6.0

Good call -- added rows for these two:

> * Addition of the LogicalType union in favor of ConvertedType enum was in 2.4.0.
> * Addition of NANOS to TimeUnit 2.6.0


I added this as a note instead of a new row, as it doesn't change what is actually written into files, instead I think it signals a future incompatible change (no longer write converted type)
> * Formal deprecation of ConvertedType in 2.9.0

content/en/docs/File Format/versions.md
alamb: If we are going to make changes to the metadata, I think new magic bytes in teh footer would be an alternate.

content/en/docs/File Format/versions.md
alamb: 🤔  It was wishful thinking on my part it seems

Upon reflection I think this statement is not true

> 2. Minor releases (e.g. `2.10.0` to `2.11.0`) may add compatible
>   features, but never incompatible ones. The minor version is not recorded in the
>   file itself.

Specifically BYTE_STREAM_SPLIT , BROTlI, LZ4 and ZSTD were added in minor releases and are forward incompatible

content/en/docs/File Format/versions.md
alamb: done in c438950d

alamb:
> A few more nits

Well, the point about releases not actually following semantic versioning is probably more than a nit 😆  I am not sure what to do about this though

etseidl:
> > A few more nits
> 
> Well, the point about releases not actually following semantic versioning is probably more than a nit 😆 I am not sure what to do about this though

Add an asterisk...
```
* Asperational. Real world versioning may differ.
```

content/en/docs/File Format/versions.md
alamb: Done in in 45b2b949

content/en/docs/File Format/versions.md
alamb: Done in 45b2b949

alamb:
> > A few more nits
> 
> Well, the point about releases not actually following semantic versioning is probably more than a nit 😆 I am not sure what to do about this though

🤔  Maybe we could roll it into a real "v3" 🤔  

Will wait a while for more comments on this PR before I open that can of worms on the mailing list

alamb:
I think I have addressed most feedback and regenerated the preview: https://alamb.github.io/parquet-site/docs/file-format/versions/

content/en/docs/File Format/versions.md
etseidl: That's one option, but I think we'll always want 'PAR', so that limits how many changes we can make with the remaining byte ('1' and 'E' are taken, so I guess that leaves 254 more, fewer if we restrict ourselves to ASCII 🤣). I proposed in the M/L augmenting the header, which would be a forwards compatible change that leaves us leeway to use anything we want to convey versioning info.

alamb: `PAR2` for maximum confusion!

etseidl (approved):
Looks good (mod SemVer). Thanks for doing this!

content/en/docs/File Format/versions.md
etseidl: You owe me a new keyboard! I just sprayed coffee all over it 🤣

content/en/docs/File Format/versions.md
emkornfield: I think it allows use to mark have richer hyper links on the other page.  For the compatibility chart, we already had to populate some of this data:

https://github.com/apache/parquet-site/blob/production/data/implementations/features/encodings.yaml#L33

emkornfield: This also means one more page necessary to update when we add something new?

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,267 @@
...
+  column values.
+* **Data Page V2 headers**: a reader that only understands `DataPageHeader`
+  cannot parse `DataPageHeaderV2` pages.
+
+[classifies changes]: https://github.com/apache/parquet-format/blob/master/CONTRIBUTING.md#compatibility-and-feature-enablement
+
+## `FileMetadata` version field
+
+Each Parquet file has a `version` field in the [`thrift FileMetadata`] that
+declares which features the file may use, and thus what a reader **must** support
```
emkornfield: I think my issue is the linear nature of versioning here it imposes on readers.  In a perfect world every reader would implement everything they needed to up as it is released.  But this means a reader needs to move in lock-step with the major version of the header.  For example, it lets say we release the following features:
1.  Backward incompatible feature that isn't strictly better for everyone (V3)
2. Awesome new encoding that a lot more people care about (V4)

By this spec, any writer would need to write V4.  This gives readers two choices:
1.  Cheat and try to read the data anyways (this makes version less useful in general, and I think one of the reasons some writer always wrote "1".  Readers were capable of reading new encodings (and pretty cleanly detecting when they couldn't so people ignored the guidance).
2. Implement both V3 and V4 before they get the benefit of V4 (this might have much longer delays given a lot of parquet implementations are volunteer driven).

Is there a way to reconcile this?

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,267 @@
...
+issues][closing-out-2.0].
+
+## `parquet-format` release versions
+
+The Thrift definition is released independently of implementations such as
+parquet-java or arrow-rs, following the Apache release process and
+[semantic versioning]:
+
+1. The major version corresponds to the [`thrift FileMetadata`] `version` field.
+2. Minor releases (e.g. `2.10.0` to `2.11.0`) may add compatible
```
emkornfield: Again, I'm not sure SemVer is really the right model here.  Partially for reasons as outlined in [the comment above](https://github.com/apache/parquet-site/pull/186/changes#r3364542047).  

But also, because its philosophy is maor version bumps should be rare.  If we really do add a handful of new encodings and a new footer it would get bumped pretty quickly.  This might naturally lead to delays similar to what we saw with v2.  From my perspective, as long as we agree we are OK bumping for every new feature I'm generally OK with this.  A second alternative is to specify a very short maximum cadence that features will collect for (e.g. 1 Month).

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,267 @@
...
+| [xxHash-based bloom filters] | [2.7.0] | [2.6.0..2.7.0] | [Approved 2019-09-09]                                     |
+| [Bloom filter length] | [2.10.0] | [2.9.0..2.10.0] |                                                           |
+| [Page index] | [2.4.0] | [2.3.1..2.4.0] |                                                           |
+| [Page CRC32 checksum] | [1.0.0] | [1.0.0][tree-1.0.0] |                                                           |
+| [Size statistics] | [2.10.0] | [2.9.0..2.10.0] | [Approved 2023-11-14]                                     |
+| [Geospatial statistics] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2025-02-09]                                     |
+| [Binary protocol extensions] | [2.11.0] | [2.10.0..2.11.0] | [Approved 2024-09-06]                                     |
+| [IEEE 754 total order and NaN counts] | not yet released | [#514] | [Approved 2026-05-26]                                     |
+| [LogicalType union] | [2.4.0] | [2.3.1..2.4.0] | Supersedes `ConvertedType` enum<br/>deprecated in [2.9.0] |
```
emkornfield, line 128: nit: I think the deprecation of converted type is not forward-compabile, you get different values if you don't understand unsigned vs signed integer types.  This could actually lead to incorrect system results.

etseidl: The deprecation just means no new values are added to the ConvertedType enum. Writers are still required to populate both ConvertedType and LogicalType (although the wording changes depending on which paragraph you read...one [place](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#metadata) says "should", two [paragraphs later](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#compatibility) it says "must").

content/en/docs/File Format/versions.md
alamb: Ok, I will update to be data driven

content/en/docs/File Format/versions.md
alamb: This is a really nice description of the core tradeoffs with using Versions (and I think one of the major points we need to resolve to  arrive at agreement about how to move versioning forward)

I am not sure there are simple ways to resolve this tradeoff and I believe there are a variety of different opinions

Thus, What I plan to do is to update this PR to *REMOVE* the a statement on semantic versioning, and keep it focused on what the current state of the spec /features is. We can then continue to have the fun (FUN!) discussion about how to version / signal new features in the corresponding mailing list thread.

content/en/docs/File Format/versions.md
emkornfield: @alamb I'm fine with this as a follow-up or we can always reconcile separately.

content/en/docs/File Format/versions.md
alamb: Clarified in d8c63bd that Parquet does NOT follow semantic versioning

content/en/docs/File Format/versions.md
alamb: Clarified in d8c63bd that Parquet does NOT follow semantic versioning

content/en/docs/File Format/versions.md
alamb: Thanks -- I did some research, and I think it would be relatively straightforward to update this page to be data driven.

Upon reflection I would like to do it as a follow on PR to minimize the diff in this PR (I think unifying the data would result in changes to more files and would be harder to track)

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,266 @@
...
+
+## `FileMetadata` version field
+
+Each Parquet file has a `version` field in the [`thrift FileMetadata`] that
+declares which features the file may use, and thus what a reader **must** support
+to read it.
+
+**Note**: Many writers set the version field to `1` even for files that use
+format 2.0 features, which has caused [confusion and interoperability
+issues][closing-out-2.0].
```
jorisvandenbossche: The first paragraph giving meaning to the "version" metadata field seems a bit confusing/misleading, together with the note, and moreover with the fact that the thrift itself specifically says this should be hardcoded to "1":

https://github.com/apache/parquet-format/blob/74001e41f5c5a1856b29be115f9c992cab16a4bf/src/main/thrift/parquet.thrift#L1365-L1374

```
...
struct FileMetaData {
  /** Version of this file
    *
    * As of December 2025, there is no agreed upon consensus of what constitutes
    * version 2 of the file. For maximum compatibility with readers, writers should
    * always populate "1" for version. For maximum compatibility with writers,
    * readers should accept "1" and "2" interchangeably.  All other versions are
    * reserved for potential future use-cases.
    */
  1: required i32 version
...
```

Or when keeping the text here, the note in the thrift file should be updated to match better?

content/en/docs/File Format/versions.md
alamb: See also some other ideas of encoding the parquet-format version in the metadata itself
- https://github.com/apache/parquet-format/pull/581
- https://github.com/apache/parquet-format/pull/582

content/en/docs/File Format/versions.md
alamb: I updated the text to say minor releases can contain forward incompatible changes

content/en/docs/File Format/versions.md
alamb: Actually, here is a proposal for how to encode the parquet-format version directly in the meatdata
- https://github.com/apache/parquet-format/pull/581
- https://github.com/apache/parquet-format/pull/582

content/en/docs/File Format/versions.md
alamb: I think you are right -- I will back this description off to match what is in the thrift

alamb: In 89159332dc770c64d88f48fcdeb24be53fc82161

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,261 @@
...
+
+## `parquet-format` release versions
+
+The Thrift definition is released independently of implementations such as
+parquet-java or arrow-rs, following the Apache release process. Note that
+release numbering **DOES NOT FOLLOW** [semantic versioning]:
+
+1. The major version corresponds to the [`thrift FileMetadata`] `version` field.
+2. Minor releases (e.g. `2.10.0` to `2.11.0`) sometimes contain forward
+   incompatible features. The minor version is not recorded in the file itself.
```
etseidl:
```suggestion
parquet-java or arrow-rs, following the Apache release process. This
release version is not recorded in the FileMetaData. Note that
release numbering **DOES NOT FOLLOW** [semantic versioning]:
minor releases (e.g. `2.10.0` to `2.11.0`) sometimes contain forward
incompatible features.
```

One last area ascribing meaning to the `version` field.

alamb:
Thank you @etseidl . I also removed the V1/V2 columns in https://github.com/apache/parquet-site/pull/186/commits/1126b539491040f97627f720f9c593a9705a4377 per @pitrou 's [suggestion on the mailing list](https://lists.apache.org/thread/7q2o02sfodz0r19gn786632fvg8d9dg3)and also regenerated the preview and screenshots.

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,260 @@
...
+parquet-java or arrow-rs, following the Apache release process. This
+release version is not recorded in the FileMetaData. Note that
+release numbering **DOES NOT FOLLOW** [semantic versioning]:
+minor releases (e.g. `2.10.0` to `2.11.0`) sometimes contain forward
+incompatible features.
+
+## Adding new features
+
+New features are added by discussion and voting on the [parquet dev mailing list]
+(full process [here]). Once approved, a feature is added to the spec and ships in
```
emkornfield:
```suggestion
(full process [contributing guide]). Once approved, a feature is added to the spec and ships in
```

emkornfield: If accepted update the link below.

emkornfield (approved):
Looking forward to capturing this in a structured way but changes look reasonable.

package.json
```diff
@@ -18,9 +18,9 @@
   "homepage": "https://github.com/apache/parquet-site#readme",
   "devDependencies": {
     "@mermaid-js/mermaid-cli": "^11.4.2",
-    "autoprefixer": "^10.4.18",
+    "autoprefixer": "^10.5.0",
```
emkornfield: are version bumps necessary?

alamb: No, I will revert

Update: reverted in 829d6cf5

alamb:
> Looking forward to capturing this in a structured way but changes look reasonable.

I will make a PR right after I merge this one to convert to using a structured format

pitrou (approved):
This is neat, thanks a lot @alamb

emkornfield:
@alamb anything we are waiting for on merging?

alamb:
> @alamb anything we are waiting for on merging?

Not that I know of -- I will plan to merge tomorrow or Friday

content/en/docs/File Format/versions.md
```diff
@@ -0,0 +1,260 @@
...
+| [ZSTD] | [2.4.0] | [2.3.1..2.4.0] |  |
+| [Modular encryption] | [2.7.0] | [2.6.0..2.7.0] | [Approved 2019-01-16] |
+
+
+> **Note:** Files with an [encrypted footer] use different magic bytes (`PARE`
+> instead of `PAR1`), making it clear to readers they must support modular
+> encryption to read the file; [plaintext footer] files use `PAR1` so legacy
+> readers can still read their unencrypted columns.
+
+## Forward compatible additions
```
wgtmac, line 113: Do we want to include the new `IEEE754TotalOrder` column order just released in 2.13.0? I think it is a compatible change if readers just regard the new column order as an unknown order and ignore its min/max stats.

wgtmac (approved):

alamb:
Soryr I am behind on this.  I will get it merged / follow on this week

content/en/docs/File Format/versions.md
alamb: Done in 9218b90a

alamb:
Thank you everyone for your help and support. I will now work on turning this into a data driven page per @emkornfield's request

content/en/docs/File Format/versions.md
emkornfield: Also https://github.com/apache/parquet-format/pull/588 to not need version info directly in metadata (although we could add it).

alamb:
The new page is live: https://parquet.apache.org/docs/file-format/versions/

alamb:
Here is a follow on PR to create these tables using data in yaml files(rather than hard coded markdown)
- https://github.com/apache/parquet-site/pull/189

PR #189, opened 2026-06-24
Make format version pages data driven

alamb:
Follow on to
- https://github.com/apache/parquet-site/pull/186

Rendered Preview: 
- https://alamb.github.io/parquet-site/docs/file-format/versions/
- https://alamb.github.io/parquet-site/docs/file-format/implementationstatus/

# Rationale
Implements @emkornfield 's request in https://github.com/apache/parquet-site/pull/186#discussion_r3363777456
> could we somehow integrate this as YAML data + rendering like we do for feature compatibility?

Among other things, rendering the page using data files forces it to remain consistent with the implementation status  page. 

# Changes

Updates the Parquet format version and implementation status pages to generate their feature tables from structured data under data/implementations.

# Planned follow on 

Making this change already shows value -- some features only on the [versions page](https://parquet.apache.org/docs/file-format/versions/)  but not the [implementation status page ](https://parquet.apache.org/docs/file-format/implementationstatus/)

This PR simply ignores rows that have `implementation_status: false` on the implementation status page

As a follow on PR, I will  propose adding the new features to the implementation status page, but this PR is already quite large

  Examples:
   - logical-type-union
   - logical-nanosecond-time-timestamp
   - logical-variant-shredding
   - format-data-page-v1
   - format-geospatial-statistics
   - format-binary-protocol-extensions
   - format-ieee-754-total-order-and-nan-counts


content/en/docs/File Format/implementationstatus.md
```diff
@@ -28,5 +28,5 @@ The value in each box means:
 * [duckdb](https://github.com/duckdb/duckdb) (C++)
 * [polars](https://github.com/pola-rs/polars) (Rust)
 
-<!-- Status source in data/implementations -->
+<!-- Data driven table, see /layouts/shortcodes -->
```
alamb, line 31: I added a readme and new link in layouts/shortcodes

content/en/docs/File Format/versions.md
```diff
@@ -72,190 +72,18 @@ the next parquet-format release.
 
 Forward incompatible features and the format version each became available in:
 
-| Feature | Released in | Source | Notes |
```
alamb, line 75: The point of the PR is to automatically derive this table from the .yml files

data/implementations/features/compressions.yaml
```diff
@@ -2,24 +2,49 @@ category_id: compressions
 features:
   - id: compression-uncompressed
     display_name: UNCOMPRESSED
-
+    compatibility: forward_incompatible
```
alamb, line 5: I added the various fields needed for the version page to the existing data files. I could probably make this less verbose by automatically generating the urls, but I think this is clearer and easier to understand

data/implementations/features/encodings.yaml
```diff
@@ -2,37 +2,67 @@ category_id: encodings
...
+    compatibility: forward_incompatible
+    released_in: 2.8.0
+    release_date: '2020-01-13'
+    feature_url: https://github.com/apache/parquet-format/blob/master/Encodings.md#byte-stream-split-byte_stream_split--9
+    source_url: https://github.com/apache/parquet-format/compare/apache-parquet-format-2.7.0...apache-parquet-format-2.8.0
+    version_note: "[Approved 2019-12-03](https://lists.apache.org/thread/xs5qt2odm299pxgqb22mty2csc1so5yr)"
   - id: encoding-byte-stream-split-extended
     display_name: BYTE_STREAM_SPLIT (Additional Types)
     spec_url: https://github.com/apache/parquet-format/commit/e517ac4dbe08d518eb5c2e58576d4c711973db94
-    format_version:
```
alamb, line 36: some features already had a "format_version" which was redundant with the information from the versions page, so I consolidated them into the same fields


From 

```yaml
    format_version:
      version: "2.11.0"
      date: "2024-03-21"
```

To 

```yaml
    released_in: 2.11.0
    release_date: '2024-03-21'
```

layouts/shortcodes/README.md
```diff
@@ -0,0 +1,15 @@
+# Hugo shortcodes
```
alamb, line 1: I found the whole hugo rendering thing new, so I also left some small documentation

etseidl:
Thanks @alamb, this is neat! One thing occurred to me while verifying external links...it would be nice for the features to link to their specific PR. The current "Source" column links to all changes between versions, so there's a lot of housekeeping changes to wade through.

Not for now, but as a follow-on.

data/implementations/features/logical-types.yaml
```diff
@@ -2,81 +2,178 @@ category_id: logical-types
...
+    version_display_name: UUID (FIXED_LEN_BYTE_ARRAY(16))
+    compatibility: forward_compatible
+    released_in: 2.6.0
+    feature_url: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#uuid
+    source_url: https://github.com/apache/parquet-format/compare/apache-parquet-format-2.5.0...apache-parquet-format-2.6.0
   - id: logical-int-types
     display_name: 8, 16, 32, 64 bit signed and unsigned INT
-
+    version_display_name: Signed and unsigned integer logical types (INT32, INT64)
+    compatibility: forward_compatible
```
emkornfield, line 24: I think this is incompatible?

data/implementations/features/logical-types.yaml
```diff
@@ -2,81 +2,178 @@ category_id: logical-types
...
     display_name: DECIMAL (INT32)
-
+    compatibility: forward_compatible
+    released_in: 2.1.0
+    feature_url: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
+    source_url: https://github.com/apache/parquet-format/compare/parquet-format-2.0.0...parquet-format-2.1.0
   - id: logical-decimal-int64
     display_name: DECIMAL (INT64)
-
+    compatibility: forward_compatible
```
emkornfield, line 34: technically incompatible.  numbers cannot be understood properly without scale?

data/implementations/features/logical-types.yaml
emkornfield: negative physical values are interpetted incorrectly?

layouts/shortcodes/format-versions.html
```diff
@@ -0,0 +1,103 @@
...
+      {{- $displayName := $feature.display_name -}}
+      {{- if $feature.version_display_name -}}
+        {{- $displayName = $feature.version_display_name -}}
+      {{- end -}}
+      {{- $featureUrl := $feature.feature_url -}}
+      {{- if and (not $featureUrl) $feature.spec_url -}}
+        {{- $featureUrl = $feature.spec_url -}}
+      {{- end -}}
+      {{- $releasedIn := printf "%v" $feature.released_in -}}
+      {{- $legacyReleaseTags := slice "1.0.0" "2.0.0" "2.1.0" -}}
```
emkornfield: what does the slice do here?

layouts/shortcodes/implementation-status.html
```diff
@@ -45,6 +46,8 @@ <h3 id="{{ $categoryId }}"><a href="#{{ $categoryId }}">{{ $category.name }}</a>
       </thead>
       <tbody>
         {{- range $features -}}
+          {{- /* Include rows by default; skip rows explicitly marked implementation_status: false. */ -}}
+          {{- if ne .implementation_status false -}}
```
emkornfield, line 52: Thank you for checking this.

emkornfield (approved):
Overall, looks good to me, I think some of the logical changes are actually forward incompatible but that might be a broader conversation on what exactly is compatible/incompatible with them.  

Thank you for doing this.

data/implementations/features/logical-types.yaml
etseidl: I think we've agreed that new logical types are forward compatible.

> New logical types are considered forward compatible despite the loss of semantic meaning.

https://github.com/apache/parquet-format/blob/master/CONTRIBUTING.md#compatibility-and-feature-enablement

emkornfield: Yeah, this is broader discussion, I think we should clarify the wording here.  we can leave it as is, but I worry about the difference between different semantics:
1.  Losing information that a int64 represents a timestamp is fine in my mind.  It still looks like an integer, and following arithmetic properties.
2. Losing information that an int64 is actually a uint64 feels like a data corruption case.

But I'd have to clarify my thinking here a bit more.  We don't need to make the change here.

data/implementations/features/encodings.yaml
emkornfield: my one concern is 2.11.0, interpretted properly as a string without quote marks?

data/implementations/features/logical-types.yaml
emkornfield: Same comment as https://github.com/apache/parquet-site/pull/189#discussion_r3470700287 applies, not a blocker.

layouts/shortcodes/format-versions.html
alamb: Apparently in hugo makes an array -- this is part of some overly complicated logic to translate release names like `1.0.0` into tag names (whose format changed over time)

I simplified the whole thing by changing this to a lookup table -- which simplifies this logic and I think makes it clearer what is going on

data/implementations/features/encodings.yaml
alamb: Good idea -- I quoted them all in commit 9f8d58a2

alamb (commented):
Thank you @etseidl  and @emkornfield 

I reworked the rendering to 
1. use a table for each parquet format release to avoid fancy rendering logic
2. restore a Notes column
3. quoted the version numbers

I have also re-generated the rendered preview https://alamb.github.io/parquet-site/docs/file-format/versions/

I think there are two follow ons here:
1. Figure out what to do with the features that are in parquet-format but don't have entry on implementation status: https://github.com/apache/parquet-site/pull/189/changes#r3470611445
2. @etseidl 's suggestion to research the approving PRs https://github.com/apache/parquet-site/pull/189#issuecomment-4793520337 (I'll do this as a follow on PR)


Stuff I am not really sure how to proceed: 
- @emkornfield 's comment in https://github.com/apache/parquet-site/pull/189/changes#r3470700287 about what actually constitutes "forward compatible" -- I think this is a larger consensus building exercise so I would like to not address it in this PR

emkornfield:
> Figure out what to do with the features that are in parquet-format but don't have entry on implementation status: https://github.com/apache/parquet-site/pull/189/changes#r3470611445

Honestly I think most of these are so old, that it doesn't necessary add value, other then maybe providing a note on that page?  Is there anything worth surfacing?

Looking more closely, these seem to be the modern ones:
 * logical-variant-shredding - should be covered under variant type support? But maybe called out separately
* format-geospatial-statistics - should be covered under geospatial support? But maybe called out separately
* format-binary-protocol-extensions - Not clear if this is actually something on its own without a defined footer we want to use?
* format-ieee-754-total-order-and-nan-counts - This should be added (and we would ideally add a 2026 feature column for read support).

> @emkornfield 's comment in https://github.com/apache/parquet-site/pull/189/changes#r3470700287 about what actually constitutes "forward compatible" -- I think this is a larger consensus building exercise so I would like to not address it in this PR

Yeah, I have a feeling this might come up in the context of versioning conversations no need to block.

alamb:
> Looking more closely, these seem to be the modern ones:

I will try and address it in a follow on PR

etseidl:
> 2\. @etseidl 's suggestion to research the approving PRs [#189 (comment)](https://github.com/apache/parquet-site/pull/189#issuecomment-4793520337) (I'll do this as a follow on PR)

I certainly didn't mean to task you 😉. I'd be happy to compile the data...I'd just need rendering help.

alamb:
> > 2. @etseidl 's suggestion to research the approving PRs [#189 (comment)](https://github.com/apache/parquet-site/pull/189#issuecomment-4793520337) (I'll do this as a follow on PR)
> 
> I certainly didn't mean to task you 😉. I'd be happy to compile the data...I'd just need rendering help.

Oh no problem -- I'll send an agent to do it and then verify. Your help will be most useful in helping to review it as well

alamb:
I think I have addressed all feedback in this PR -- so I will plan to merge it in tomorrow unless people would like more time or have other comments. I'll also file tickets to track follow on work

alamb:
> > > 2. @etseidl 's suggestion to research the approving PRs [#189 (comment)](https://github.com/apache/parquet-site/pull/189#issuecomment-4793520337) (I'll do this as a follow on PR)
> > 
> > 
> > I certainly didn't mean to task you 😉. I'd be happy to compile the data...I'd just need rendering help.
> 
> Oh no problem -- I'll send an agent to do it and then verify. Your help will be most useful in helping to review it as well

- filed a ticket https://github.com/apache/parquet-site/issues/191

alamb:
> > Looking more closely, these seem to be the modern ones:
> 
> I will try and address it in a follow on PR

Tracking in 
- https://github.com/apache/parquet-site/issues/192

alamb:
@etseidl  here is a PR to add the PR research
- https://github.com/apache/parquet-site/pull/193

PR #197, opened 2026-08-05
Explicitly use an arithmetic right shift in IEEE754 total order

HuaHuaY:
For signed integers, the `>>` operator does not necessarily represent an arithmetic right shift in some languages ​​(for example, in C++17, it is implementation-defined). A comment is included to serve as a reminder.


alamb (approved):
Makes sense to me (though I do think Rust, the language the example is written in, does explicitly defines this as an arithmetic operation: https://doc.rust-lang.org/stable/reference/expressions/operator-expr.html#arithmetic-and-logical-binary-operators)

alamb:
Thank you @HuaHuaY

wgtmac (approved):

wgtmac:
Thanks @HuaHuaY and @alamb!

